Issue #1, opened 2023-02-09
Missing ip and external_hostname making mh incompatible with multihost that it should replace.


jakub-vavra-cz:
As a replacement for pytest-multihost the host ip and external_hostname should be implemented and supported.
The expectation that (internal) hostname is resolvable is not true for machines in cloud and even for use with local containers it would be better to be able to define and use ip addresses.
The ssh logic should be to use first ip (if present), then external_hostname (if present) and then hostname.

pbrezina:
Thank you. I implemented it little bit differently and split ssh related options to standalone key.

See: https://github.com/next-actions/pytest-mh/commit/7507489beeeeaf25c08218402c09a4e89c546352

PR #2, opened 2023-02-13
Add support for ip address.

jakub-vavra-cz:
Partial fix for issue #1 by adding support for an ip address.


pbrezina:
This was solved by https://github.com/next-actions/pytest-mh/commit/7507489beeeeaf25c08218402c09a4e89c546352.

PR #4, opened 2023-06-29
ssh: Respect the user and password provided by the caller

aplopez:
This piece of code fails because the user and password provided by the caller are not used.
```
    with client.ssh('ci', 'Secret123') as ssh:
        result = ssh.run('whoami')
        assert result.rc == 0
        assert result.stdout == 'ci'
```
This commit fixed that using them whenever they are available.


pbrezina:
Hi, thank you. I think this is already fixed by https://github.com/next-actions/pytest-mh/commit/805efa818ccaef7dd69449c2dea532365bc0c836

By the way, user and password can only be string not None as said in the method declaration.

PR #5, opened 2023-07-19
fs: wc command added

patriki01:
Simple PR adding wc command with arguments.

### DIFF
--- a/pytest_mh/utils/fs.py
+++ b/pytest_mh/utils/fs.py
@@ -4,7 +4,7 @@ import base64
 import textwrap
 
 from .. import MultihostHost, MultihostUtility
-from ..ssh import SSHLog
+from ..ssh import SSHLog, SSHProcessResult
 
 __all__ = ["LinuxFileSystem"]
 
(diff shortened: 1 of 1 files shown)


pytest_mh/utils/fs.py
```diff
@@ -442,3 +442,32 @@ def __gen_chattrs(
...
+        :param word: Print the word counts, defaults to False
+        :type word: bool, optional
+        :param bytes: Print the byte counts, defaults to False
+        :type bytes: bool, optional
+        :return: Result of process
+        :rtype: SSHProcessResult
+        """
+        args = []
+        if lines:
+            args = ["-l"]
```
pbrezina, line 465: args.append here as well

pytest_mh/utils/fs.py
```diff
@@ -442,3 +442,32 @@ def __gen_chattrs(
             cmds.append(f"chgrp '{group}' '{path}'")
 
         return " && ".join(cmds)
+
+    def wc(self, file: str, lines: bool = False, word: bool = False, bytes: bool = False) -> SSHProcessResult:
```
pbrezina, line 446: There is also -m chars.

pbrezina (changes requested):

pytest_mh/utils/fs.py
patriki01: added

pytest_mh/utils/fs.py
patriki01: fixed

pbrezina:
Ack.

PR #7, opened 2023-08-05
fs: add append method

aborah-sudo:
Add append support for fs module

### DIFF
--- a/pytest_mh/utils/fs.py
+++ b/pytest_mh/utils/fs.py
@@ -207,6 +207,38 @@ class LinuxFileSystem(MultihostUtility):
             log_level=SSHLog.Error,
         )
 
+    def append(
+        self,
+        path: str,
+        contents: str,
+        *,
+        dedent: bool = True,
+    ) -> None:
+        """
+        Append to a remote file.
+
+        :param path: File path.
+        :type path: str
+        :param contents: File contents to write.
+        :type contents: str
+        :param dedent: Automatically dedent and strip file contents, defaults to True
+        :type dedent: bool, optional
+        """
+        if dedent:
+            contents = textwrap.dedent(contents).strip()
+
+        self.backup(path)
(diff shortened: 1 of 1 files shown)


pytest_mh/utils/fs.py
```diff
@@ -207,6 +207,39 @@ def write(
...
+            contents = textwrap.dedent(contents).strip()
+
+        self.backup(path)
+        self.logger.info(f'Writing file "{path}" on {self.host.hostname}', extra={"data": {"Contents": contents}})
+
+        self.host.ssh.run(
+            f"""
+                set -ex
+                cat >> '{path}'
+                {self.__gen_chattrs(path)}
```
pbrezina, line 237: Since we do not set mode, user and  group we do not need this line.

pytest_mh/utils/fs.py
```diff
@@ -207,6 +207,39 @@ def write(
...
+        :param contents: File contents to write.
+        :type contents: str
+        :param dedent: Automatically dedent and strip file contents, defaults to True
+        :type dedent: bool, optional
+        """
+        if dedent:
+            contents = textwrap.dedent(contents).strip()
+
+        self.backup(path)
+        self.logger.info(f'Writing file "{path}" on {self.host.hostname}', extra={"data": {"Contents": contents}})
```
pbrezina, line 231: Appending to file ...?

pbrezina (changes requested):
Thank you, see small comment inside.

Also please change the commit message to "fs: add append method"

pytest_mh/utils/fs.py
aborah-sudo: Fixed

pytest_mh/utils/fs.py
aborah-sudo: Fixed

pbrezina (approved):

PR #8, opened 2023-08-08
markers: add require marker

pbrezina:
Add additional runtime requirement, skip test if requirement is not met.

```python
@pytest.mark.require(lambda client: "files-provider" in client.features)
```

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -96,5 +96,6 @@ What does the framework do?
    config
    topology
    classes
+   runtime-requirements
    pytest
    api
--- a/docs/runtime-requirements.rst
+++ b/docs/runtime-requirements.rst
@@ -1,0 +1,54 @@
+Additional runtime requirements
+###############################
+
+Sometimes, topology itself is not enough to detect if the test can or can not
+be run and you want to check for a runtime requirement like that your program
+was built with certain configure flags or features.
+
+This can be achieved with ``pytest.mark.require(condition[, reason])`` marker
+that takes a function as a parameter and the test is skipped if the function
+returns ``False`` (the requirement was not met).
+
+The function takes all fixtures that are available to the test as parameters.
+
+.. note::
+
+    The function can either return ``bool`` or ``tuple[bool, str]``. In this
+    case, the second value is the reason for skipping the test.
+
+.. code-block:: python
+
+    @pytest.mark.topology(KnownTopology.LDAP)
+    @pytest.mark.require(
+        lambda client: "files-provider" in client.features,
+        "SSSD was not built with files provider"
+    )
+    def test_example_explicit_reason(client: Client, ldap: LDAP):
+        pass
+
+    @pytest.mark.topology(KnownTopology.LDAP)
+    @pytest.mark.require(
+        lambda client: ("files-provider" in client.features, "SSSD was not built with files provider")
+    )
+    def test_example_reason_as_tuple(client: Client, ldap: LDAP):
+        pass
+
+.. note::
+
+    The requirement is evaluated when the test is executed but before setup
+    phase, so no setup method was called on any multihost role in order to make
+    the skip fast.
+
+    If you require to setup the role, you can always call the setup method
+    directly from the function passed to the ``require`` marker.
+
+.. warning::
+
+    ``pytest-mh`` provides the ``requirement`` marker as a generic way to skip
+    a test when a condition is not met. The condition can use multihost roles
+    or other pytest fixtures used by the marked test and it can also call
+    commands on remote hosts.
+
+    The example above shows a check if an SSSD project was built with
+    "files-provider" feature, however feature detection is not part of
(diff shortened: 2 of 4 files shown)


justin-stephenson:
Hi @pbrezina, how can I go about testing this PR ?

ikerexxe:
I like the idea behind this feature but I'm not sure about its current implementation. Does it only check sssd compilation flags? Or can we check also other packages?

pbrezina:
> Hi @pbrezina, how can I go about testing this PR ?

Make sure you have this PR installed via pypi and then you can test it inside SSSD tests.

> I like the idea behind this feature but I'm not sure about its current implementation. Does it only check sssd compilation flags? Or can we check also other packages?
> 
> Let me put it in another way, passkey feature will be enabled in sssd in all fedora versions for our CI, but IPA won't. I'd like to find an easy way to only test this feature in those fedora versions where both packages have passkey feature enabled (f39+). Would it be feasible? I don't mind it it's just marker stating the minimum fedora version.

The SSSD example is just an example. It is not even implemented in this pull request. Our specific implementation of feature detection is part of sssd-test-framework, see https://github.com/SSSD/sssd-test-framework/pull/30.

This adds generic requirement marker that takes a function and provides test fixtures as parameters and you can do anything you want with it.

pbrezina:
For passkey, we need to teach the framework to detect passkey support on sssd and ipa and then we can use:

```python
@pytest.mark.require(lambda client, ipa: "passkey" in client.features and "passkey" in ipa.features)
```

Again client.features and ipa.features are part of the sssd-test-framework. And since SSSD does not support printing configuration flags, the detection is not based on this.

ikerexxe:
> Again client.features and ipa.features are part of the sssd-test-framework. And since SSSD does not support printing configuration flags, the detection is not based on this.

So, this is the interface that we'll use to define those requirements. The actual recognition of those requirements happens in sssd-test-framework. Right?

For testing purposes I created the following `requirements.txt` file:
```
pytest                                                                           
git+https://github.com/pbrezina/pytest-mh@require                                
git+https://github.com/next-actions/pytest-ticket                                
git+https://github.com/next-actions/pytest-tier                                  
git+https://github.com/ikerexxe/sssd-test-framework@test_require
```

The last line points to a branch where I removed the requirement in `sssd-test-framework` for `pytest-mh > 1.0.3`.

But I'm still missing something because I get the following error:
```
>   @pytest.mark.require(lambda client: "files-provider" in client.features)
E   AttributeError: 'Client' object has no attribute 'features'
```

pbrezina:
> > Again client.features and ipa.features are part of the sssd-test-framework. And since SSSD does not support printing configuration flags, the detection is not based on this.
> 
> So, this is the interface that we'll use to define those requirements. The actual recognition of those requirements happens in sssd-test-framework. Right?

Exactly. This is generic method to skip the test if the condition is not met. The condition itself must be implemented in the project specific extension (sssd-test-framework in our case), pytest-mh can not know how to detect particular features or other parameters.

> For testing purposes I created the following `requirements.txt` file:
> 
> ```
> pytest                                                                           
> git+https://github.com/pbrezina/pytest-mh@require                                
> git+https://github.com/next-actions/pytest-ticket                                
> git+https://github.com/next-actions/pytest-tier                                  
> git+https://github.com/ikerexxe/sssd-test-framework@test_require
> ```
> 
> The last line points to a branch where I removed the requirement in `sssd-test-framework` for `pytest-mh > 1.0.3`.
> 
> But I'm still missing something because I get the following error:
> 
> ```
> >   @pytest.mark.require(lambda client: "files-provider" in client.features)
> E   AttributeError: 'Client' object has no attribute 'features'
> ```

The features attribute is implemented in https://github.com/SSSD/sssd-test-framework/pull/30 so you need to check this out as well. It also adds a syntax sugar `@pytest.mark.isbuiltwith("files-provider")` that translates to `@pytest.mark.require(lambda client: "files-provider" in client.features)`.

https://github.com/SSSD/sssd-test-framework/pull/30 depends on this PR.

ikerexxe (approved):
LGTM!

@justin-stephenson if you are still interested in testing this PR you can use the requirements file that I indicated previously. I updated the `sssd-test-framework` branch to include Pavel's changes.

justin-stephenson:
> Add additional runtime requirement, skip test if requirement is not met.
> 
> ```python
> @pytest.mark.require(lambda client: "files-provider" in client.features)
> ```

I noticed if you change `"files-provider"` to some random string, then the feature is not found so pytest skips the test as expected however IMO it would be nice if somehow we could error out when a non-existent feature is provided as a string here. At the least we should document all available "client.features" so people don't have to look them up in the code.

pbrezina:
I added support to partial functions and note that feature detection is project specific and not implemented in pytest-mh.

justin-stephenson (approved):
Ack, Thank you.

ikerexxe (approved):
LGTM. Thanks for the patch!

PR #11, opened 2023-08-16
Add back 'markers: add require marker'

justin-stephenson:
This should be merged alongside https://github.com/SSSD/sssd-test-framework/pull/35

Add additional runtime requirement, skip test if requirement is not met.

```
@pytest.mark.require(lambda client: "files-provider" in client.features)
```

(cherry picked from commit 00169365dc6653ede0fd3a2738d3820759add354)


ikerexxe (approved):
LGTM. Thanks for taking care of it.

pbrezina:
I removed the revert commit as it did not have to be reverted in the first place and force pushed.

PR #14, opened 2023-08-24
multihost: artifacts collection fix

justin-stephenson:
Teardown of <sssd_test_framework.hosts.client.ClientHost object at 0x7f461e9f79d0> can happen which triggers restoring backup data before teardown of sssd_test_framework.roles.client.Client object occurs to collect artifacts.

### DIFF
--- a/pytest_mh/_private/fixtures.py
+++ b/pytest_mh/_private/fixtures.py
@@ -135,7 +135,7 @@ class MultihostFixture(object):
         roles: list[MultihostRole] = [x for x in self._paths.values() if isinstance(x, MultihostRole)]
         hosts: list[MultihostHost] = [x.host for x in roles]
 
-        return list(set([*hosts, *roles]))
+        return [*hosts, *roles]
 
     def _skip(self) -> bool:
         if self.data.topology_mark is None:


alexey-tikhonov (approved):
Addition of this patch helped to get correct logs in https://github.com/SSSD/sssd/pull/6873

ikerexxe:
LGTM, but CI is failing:
```
pytest_mh/_private/fixtures.py:179: error: Incompatible types in assignment (expression has type "MultihostHost[Any] | MultihostRole[Any]", variable has type "MultihostRole[Any]")  [assignment]
```

justin-stephenson:
> LGTM, but CI is failing:
> 
> ```
> pytest_mh/_private/fixtures.py:179: error: Incompatible types in assignment (expression has type "MultihostHost[Any] | MultihostRole[Any]", variable has type "MultihostRole[Any]")  [assignment]
> ```

I'm not sure what causes this error.

pbrezina:
> > LGTM, but CI is failing:
> > ```
> > pytest_mh/_private/fixtures.py:179: error: Incompatible types in assignment (expression has type "MultihostHost[Any] | MultihostRole[Any]", variable has type "MultihostRole[Any]")  [assignment]
> > ```
> 
> I'm not sure what causes this error.

Because you are reusing `item` with two different types (first with self._roles items and the with self._hosts_and_roles).

Assigning self._host and self._roles as a side effect of another property is a bad practice. You should move it to its own properties. Or avoid it at all since it is used only on one place so it can be constructed directly in the for cycle.

But I wonder, how did we reached to this state? Hosts are teared down after roles, it can not really happen that ClientHost was teared down before Client role. Or am I missing something?

justin-stephenson:
> Because you are reusing `item` with two different types (first with self._roles items and the with self._hosts_and_roles).
> 
> Assigning self._host and self._roles as a side effect of another property is a bad practice. You should move it to its own properties. Or avoid it at all since it is used only on one place so it can be constructed directly in the for cycle.

Okay thanks I will address this.
> 
> But I wonder, how did we reached to this state? Hosts are teared down after roles, it can not really happen that ClientHost was teared down before Client role. Or am I missing something?

Hosts are being tore down before roles, see the mh.log where I added logging the item inside the `for item in reversed(self._hosts_and_roles):` loop. 
[mh.log](https://github.com/next-actions/pytest-mh/files/12464771/mh.log)

pbrezina:
> > Because you are reusing `item` with two different types (first with self._roles items and the with self._hosts_and_roles).
> > Assigning self._host and self._roles as a side effect of another property is a bad practice. You should move it to its own properties. Or avoid it at all since it is used only on one place so it can be constructed directly in the for cycle.
> 
> Okay thanks I will address this.
> 
> > But I wonder, how did we reached to this state? Hosts are teared down after roles, it can not really happen that ClientHost was teared down before Client role. Or am I missing something?
> 
> Hosts are being tore down before roles, see the mh.log where I added logging the item inside the `for item in reversed(self._hosts_and_roles):` loop. [mh.log](https://github.com/next-actions/pytest-mh/files/12464771/mh.log)

But why? The intention is to do roles first, of course. We have reverse(*hosts, *roles) so roles comes first in the for cycle. How did we get to the state when it is mixed up.

justin-stephenson:
> 
> But why? The intention is to do roles first, of course. We have reverse(*hosts, *roles) so roles comes first in the for cycle. How did we get to the state when it is mixed up.

Okay I wasn't clear that was the intention, I see now the following patch fixes this issue due to `set` being unordered. I will update the PR if you agree.
~~~
diff --git a/pytest_mh/_private/fixtures.py b/pytest_mh/_private/fixtures.py
index f2a6649..1938be6 100644
--- a/pytest_mh/_private/fixtures.py
+++ b/pytest_mh/_private/fixtures.py
@@ -132,7 +132,7 @@ class MultihostFixture(object):
         roles: list[MultihostRole] = [x for x in self._paths.values() if isinstance(x, MultihostRole)]
         hosts: list[MultihostHost] = [x.host for x in roles]
 
-        return list(set([*hosts, *roles]))
+        return list([*hosts, *roles])
 
     def _setup(self) -> None:
         """
~~~

pbrezina:
Ah, ok. The purpose of set was to remove any potential duplicates, but I don't think that can really happen. So simply returning `[*hosts, *roles]` (without the list constructor) will suffice.

ikerexxe (approved):
LGTM. Thanks for taking care of it

pbrezina (approved):

PR #17, opened 2023-09-19
fs: add chmod utility

patriki01:
Added method to perform chmod.
Feel free to suggest changes. :robot:

### DIFF
--- a/pytest_mh/utils/fs.py
+++ b/pytest_mh/utils/fs.py
@@ -527,3 +527,20 @@ class LinuxFileSystem(MultihostUtility):
             args.append("-m")
 
         return self.host.ssh.exec(["wc", *args, file])
+
+    def chmod(self, mode: str, path: str, args: list[str] = []) -> SSHProcessResult:
+        """
+        Change file/folder mode bits.
+        Mode can be specified in two ways: octal number e.g. "666", "444" or
+        a symbolic representation of changes e.g. "u=rw,go=r", "go-rw"
+
+        :param mode: New mode of file/folder
+        :type mode: str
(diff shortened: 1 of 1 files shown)


pytest_mh/utils/fs.py
```diff
@@ -527,3 +527,20 @@ def wc(
...
+
+    def chmod(self, mode: str, path: str, args: list[str] = []) -> SSHProcessResult:
+        """
+        Change file mode bits.
+        Mode can be specified in two ways: octal number e.g. "666", "444" or
+        a symbolic representation of changes e.g. "u=rw,go=r", "go-rw"
+
+        :param mode: New mode of file
+        :type mode: str
+        :param path: Folder whose permissions change
```
ikerexxe, line 539: This could also be a file, so rephrase the comment so that it's more generic or add `file` to it

pytest_mh/utils/fs.py
```diff
@@ -527,3 +527,20 @@ def wc(
...
+        :type path: str
+        :param args: Additional args
```
ikerexxe, line 541: `Additional options`? I'm not completely sure but it sounds better than `args`

ikerexxe (commented):
I have one major concern. `__gen_chattrs()` from the LinuxFileSystem class already provides a similar functionality. It's true though that it isn't as complete as this one. @pbrezina what should be done in this case?

pytest_mh/utils/fs.py
patriki01: I agree. It sounds better

patriki01:
I didn't notice that. Let's wait for Pavel then.

PR #20, opened 2023-09-26
markers: pass all defined fixtures to **kwargs in require marker

pbrezina:
Previously, we only passed the ones that are in the test function
parameters but this was unnecessarily limitting.

Now we pass all fixtures defined for the topology combined with all
other fixtures available to the function.


ikerexxe (approved):
LGTM. Thanks for the patch and for all your help getting this done!

Issue #16, opened 2023-09-14
Add utility to slow down the connection


ikerexxe:
Use `tc` for that purpose.

Link: https://jvns.ca/blog/2017/04/01/slow-down-your-internet-with-tc/

PR #22, opened 2023-10-05
utils: perform traffic control operations

ikerexxe:
New utility to perform traffic control operations such as slowing down the connection.

Resolves: https://github.com/next-actions/pytest-mh/issues/16

Depends on https://github.com/SSSD/sssd-ci-containers/pull/62

### DIFF
--- a/pytest_mh/utils/tc.py
+++ b/pytest_mh/utils/tc.py
@@ -1,0 +1,95 @@
+from __future__ import annotations
+
+from .. import MultihostHost, MultihostUtility
+from ..ssh import SSHLog
+
+__all__ = ["LinuxTrafficControl"]
+
+
+class LinuxTrafficControl(MultihostUtility):
+    """
+    Perform traffic control operations on remote host.
+
+    All changes are automatically reverted when a test is finished.
+    """
+
+    def __init__(self, host: MultihostHost) -> None:
+        """
+        :param host: Remote host instance.
+        :type host: MultihostHost
+        """
+        super().__init__(host)
+        self.__interfaces: set[str] = set()
+        self.__restore: set[str] = set()
+        self.__band: int = 2
+
+    def setup_when_used(self):
+        """
+        Setup traffic control configuration.
+
+        :meta private:
+        """
+        super().setup_when_used()
+
+        # Let's find out the available interfaces
+        result = self.host.ssh.run("ip -o address", log_level=SSHLog.Error)
+        split_result = result.stdout.splitlines()
+
+        for line in split_result:
+            interface = line.split(" ")[1]
+            if interface != "lo":
+                self.__interfaces.add(interface)
+
+        commands = "set -e\n"
+        for interface in self.__interfaces:
+            commands += (
+                f"tc qdisc add dev {interface} root handle 1: prio bands 16 priomap 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0\n"
+            )
+            self.__restore.add(f"tc qdisc del dev {interface} root")
+
+        self.host.ssh.run(commands, log_level=SSHLog.Error)
+
+    def teardown_when_used(self):
+        """
+        Revert all traffic control changes.
+
+        :meta private:
+        """
+        tear: str = ""
+        for iter in self.__restore:
+            tear += iter + "\n"
+        self.host.ssh.run(tear, log_level=SSHLog.Error)
+        super().teardown_when_used()
+
+    def add_delay(self, hostname: str, time: str | int):
+        """
+        Add delay to the network connection. A maximum of 15 connections can be delayed at a time. It is recommended
+        to specify the delay from minimum to maximum to avoid starvation.
+
+        :param hostname: Target hostname.
+        :type hostname: str
+        :param time: Delay. Units can be specified; if not specified, the default is milliseconds.
+        :type time: str | int
+        """
+        ips = self.host.ssh.run(f"dig +short {hostname}", log_level=SSHLog.Error)
+        ip_list = ips.stdout.splitlines()
+
+        if isinstance(time, int):
+            time_unit = f"{time}ms"
+        else:
+            time_unit = time
+
+        commands = "set -e\n"
+        for interface in self.__interfaces:
+            commands += (
+                f"tc qdisc add dev {interface} parent 1:{self.__band} handle {self.__band*10}: netem "
+                f"delay {time_unit}\n"
+            )
+            for ip in ip_list:
+                commands += (
+                    f"tc filter add dev {interface} protocol ip parent 1:0 u32 match ip dst {ip} "
+                    f"flowid 1:{self.__band}\n"
+                )
+
+        self.host.ssh.run(commands, log_level=SSHLog.Error)
+        self.__band += 1
--- a/pytest_mh/utils/tc.py
+++ b/pytest_mh/utils/tc.py
@@ -20,7 +20,8 @@ class LinuxTrafficControl(MultihostUtility):
         """
         super().__init__(host)
         self.__interfaces: set[str] = set()
-        self.__restore: set[str] = set()
+        self.__restore_root: set[str] = set()
+        self.__restore_filters: dict[str, int] = dict()
         self.__band: int = 2
 
     def setup_when_used(self):
@@ -45,7 +46,7 @@ class LinuxTrafficControl(MultihostUtility):
             commands += (
                 f"tc qdisc add dev {interface} root handle 1: prio bands 16 priomap 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0\n"
             )
-            self.__restore.add(f"tc qdisc del dev {interface} root")
+            self.__restore_root.add(f"tc qdisc del dev {interface} root")
 
         self.host.ssh.run(commands, log_level=SSHLog.Error)
 
@@ -56,7 +57,7 @@ class LinuxTrafficControl(MultihostUtility):
         :meta private:
         """
         tear: str = ""
-        for iter in self.__restore:
+        for iter in self.__restore_root:
             tear += iter + "\n"
         self.host.ssh.run(tear, log_level=SSHLog.Error)
         super().teardown_when_used()
@@ -87,9 +88,34 @@ class LinuxTrafficControl(MultihostUtility):
             )
             for ip in ip_list:
                 commands += (
-                    f"tc filter add dev {interface} protocol ip parent 1:0 u32 match ip dst {ip} "
+                    f"tc filter add dev {interface} protocol ip prio {self.__band} parent 1:0 u32 match ip dst {ip} "
                     f"flowid 1:{self.__band}\n"
                 )
+                self.__restore_filters[ip] = self.__band
 
         self.host.ssh.run(commands, log_level=SSHLog.Error)
         self.__band += 1
+
+    def remove_delay(self, hostname: str):
+        """
+        Remove delay in the network connection.
+
+        :param hostname: Target hostname.
+        :type hostname: str
+        """
+        ips = self.host.ssh.run(f"dig +short {hostname}", log_level=SSHLog.Error)
+        ip_list = ips.stdout.splitlines()
+
+        bands: set[int] = set()
+        for ip in ip_list:
+            if ip in self.__restore_filters:
+                bands.add(self.__restore_filters[ip])
+
+        commands = "set -e\n"
+        for interface in self.__interfaces:
+            for band in bands:
+                commands += f"tc filter del dev {interface} prio {band}\n"
+
+            commands += f"tc qdisc del dev {interface} parent 1:{band} handle {band*10}: netem\n"
+
+        self.host.ssh.run(commands, log_level=SSHLog.Error)


pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,54 @@
...
+    All changes are automatically reverted when a test is finished.
+    """
+
+    def __init__(self, host: MultihostHost) -> None:
+        """
+        :param host: Remote host instance.
+        :type host: MultihostHost
+        """
+        super().__init__(host)
+        self.__restore = set()  # type: set[str]
```
pbrezina, line 22: `self.__restore: set[str] = set()`

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,54 @@
...
+
+        # Let's find out the interface that connects the systems
+        self.__interface: str = ""
+        result = self.host.ssh.run(f"dig +short {host.hostname}", log_level=SSHLog.Error)
+        try:
+            result = self.host.ssh.run(f"ip -o address | grep {result.stdout}", log_level=SSHLog.Error)
+            self.__interface = result.stdout.split(" ")[1]
+        except SSHProcessError:
+            print(f"{host.hostname} doesn't provide 'ip' command")
```
pbrezina, line 31: ...and what if the  command is present but fails?

I would avoid this, in the end, if a command is not available it will be visible in the stderr log of the result.

Also you can't use print, you need to use self.logger to display messages.

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        :param host: Remote host instance.
+        :type host: MultihostHost
+        """
+        super().__init__(host)
+        self.__restore = set()  # type: set[str]
+
+        # Let's find out the interface that connects the systems
+        self.__interface: str = ""
+        result = self.host.ssh.run(f"dig +short {host.hostname}", log_level=SSHLog.Error)
```
pbrezina, line 26: This all discovery should go to `setup_when_used`

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,54 @@
...
+        try:
+            result = self.host.ssh.run(f"ip -o address | grep {result.stdout}", log_level=SSHLog.Error)
+            self.__interface = result.stdout.split(" ")[1]
+        except SSHProcessError:
+            print(f"{host.hostname} doesn't provide 'ip' command")
+
+    def teardown_when_used(self):
+        """
+        Revert all firewall changes.
```
pbrezina, line 35: firewall? :-)

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,54 @@
...
+
+        :meta private:
+        """
+        for iter in self.__restore:
+            self.host.ssh.run(iter, log_level=SSHLog.Error)
```
pbrezina, line 40: It would be better to run all commands at once to avoid running multiple ssh channels.

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(self, host: MultihostHost) -> None:
+        """
+        :param host: Remote host instance.
+        :type host: MultihostHost
+        """
+        super().__init__(host)
+        self.__restore = set()  # type: set[str]
+
+        # Let's find out the interface that connects the systems
+        self.__interface: str = ""
```
pbrezina, line 25: I'd make this public, and use it as a default interface for tc command, not the only one.

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        Revert all firewall changes.
+
+        :meta private:
+        """
+        for iter in self.__restore:
+            self.host.ssh.run(iter, log_level=SSHLog.Error)
+        super().teardown_when_used()
+
+    def add_delay(self, time: int):
```
pbrezina, line 43: Add interface to parameters and use self.interface as a default.

pbrezina (changes requested):
Hi Iker, thank you. See some comments inside.

pytest_mh/utils/tc.py
ikerexxe: Ops! Sorry for the `print()` that was unintended. Can we have an empty `except` in python? Maybe using `pass`?

pytest_mh/utils/tc.py
ikerexxe: Fixed

pytest_mh/utils/tc.py
ikerexxe: Fixed

pytest_mh/utils/tc.py
ikerexxe: Fixed

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,69 @@
...
+
+        :meta private:
+        """
+        tear: str = ""
+        for iter in self.__restore:
+            tear += iter + ";"
+        self.host.ssh.run(tear, log_level=SSHLog.Error)
+        super().teardown_when_used()
+
+    def add_delay(self, time: int, interface: str = ""):
```
pbrezina, line 54: Using `None` for default value would be a better choice, it is more common in Python.

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,69 @@
...
+
+        :param time: Delay in milliseconds.
+        :type time: int
+        :param interface: Network interface.
+        :type time: ""
+        """
+        if not interface and self.interface is not None:
+            interface = self.interface
+        elif not interface and self.interface is None:
+            raise ValueError(f"Unable to detect the interface for {self.host.hostname}")
```
pbrezina, line 66: This can be simplified:

```python
if not interface and self.interface is not None:
    interface = self.interface

if not interface:
   raise ...
```

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,69 @@
...
+        :meta private:
+        """
+        super().setup_when_used()
+
+        # Let's find out the interface that connects the systems
+        result = self.host.ssh.run(f"dig +short {self.__host.hostname}", log_level=SSHLog.Error)
+        try:
+            result = self.host.ssh.run(f"ip -o address | grep {result.stdout}", log_level=SSHLog.Error)
+            self.interface = result.stdout.split(" ")[1]
+        except SSHProcessError:
```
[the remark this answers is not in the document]
pbrezina: Is any error ok? Don't you want just check for the missing command?

You would be worth to log "ip command is missing, default interface is not set" kind of message.

dig may return multiple ip addresses, how do we handle that?

pbrezina (changes requested):

pytest_mh/utils/tc.py
ikerexxe: What about this one? Should I remove the `try` or keep it as it is now?

My knowledge of Python is low and I'm trying to take advantage of what I know for Java exceptions. There if an exception is thrown you are forced to catch it, but I don't know if the same applies in Python.

ikerexxe: Sorry, I just read the next comment. Disregard the previous message.

pytest_mh/utils/tc.py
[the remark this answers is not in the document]
ikerexxe: > Is any error ok? Don't you want just check for the missing command?

Yes, how can I do that?

> dig may return multiple ip addresses, how do we handle that?

I'd default to the first one provided to dig, which is what I'm currently doing. If the test cases wants something else they can provide their selection in the interface when calling `add_delay()`.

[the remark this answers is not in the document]
pbrezina: > > Is any error ok? Don't you want just check for the missing command?
> 
> Yes, how can I do that?

bash returns 127 and `bash: a: command not found`, we could check for this. But other shell may give other rc and the string may be translated. I think that we can really just omit catching the exception. If an error occurs, its reason will be fully visible in the logs.

> > dig may return multiple ip addresses, how do we handle that?
> 
> I'd default to the first one provided to dig, which is what I'm currently doing. If the test cases wants something else they can provide their selection in the interface when calling `add_delay()`.

I need to check how tc works exactly. Ideally, we don't want to slow down ssh connection. And the use case is to slow down connection to other host. Maybe the code in setup should be available as a method to return interface for specific hostname? And probably list of interfaces would be a better choice, because SSSD can pick any available IP address, not just the first one. So in case there are multiple IP addresses, SSSD would be slow only on one of them.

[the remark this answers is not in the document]
ikerexxe: > I need to check how tc works exactly. Ideally, we don't want to slow down ssh connection. And the use case is to slow down connection to other host. Maybe the code in setup should be available as a method to return interface for specific hostname? And probably list of interfaces would be a better choice, because SSSD can pick any available IP address, not just the first one. So in case there are multiple IP addresses, SSSD would be slow only on one of them.

IIUC, your proposal would be to create a method to return the available interfaces to a host, and let each test decide which one to choose. Right? That seems problematic in the sense that the person writing the test case may not know which interface to choose.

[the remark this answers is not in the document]
pbrezina: something in the sense `delay(hostname, time)` instead of `delay(interface, time)` (or actually support both)

I need to get familiar with tc before making more suggestions. Quick search show that it should be possible [to delay only traffic to specific IP](https://serverfault.com/questions/389290/using-tc-to-delay-packets-to-only-a-single-ip-address). So `delay(hostname, time)` should work like:

```
devs = []
list_of_ips = dig hostname
for ip in list_of_ips:
     devs += (ip, dev) // find dev for ip
for ip, dev in devs
    tc ... delay communication to ip on dev by time
```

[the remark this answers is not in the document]
ikerexxe: It's better explained here https://dev.to/yugabyte/testing-an-application-with-a-network-delay-13eh, or at least I understood it better.

I'm working on a solution to this.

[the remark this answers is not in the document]
ikerexxe: Taking into account the use case I change the interface to `add_delay(hostname, time)`. The old interface doesn't make sense for that use case and it could be problematic. If at some point we need it we can add it.

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def teardown_when_used(self):
+        """
+        Revert all traffic control changes.
+
+        :meta private:
+        """
+        tear: str = ""
+        for iter in self.__restore:
+            tear += iter + ";"
```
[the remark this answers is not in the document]
ikerexxe: Done

ikerexxe:
It's ready for another round of reviews

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        ip = self.host.ssh.run(f"dig +short {hostname}", log_level=SSHLog.Error)
+
+        for interface in self.__interfaces:
+            self.host.ssh.run(
+                f"""
+                set -e
+                tc qdisc add dev {interface} root handle 1: prio
+                tc qdisc add dev {interface} parent 1:3 handle 30: netem delay {time}
+                tc filter add dev {interface} protocol ip parent 1:0 priority 3 u32 match ip dst {ip.stdout} flowid 1:3
```
pbrezina, line 71: This works. But it is not flexible, because it allows you to only add single delay. Also this delay does not apply only to the particular IP address but also any other packet which TOS sends it to 1:3 band of the priority queue.

So, we need to do some changes.

* We need to be handle-id aware, we can't just use hardcoded values.
* We must be able to create the prio root qdisc separately, because we need to set the number of bands to support. The default value is 3, but that can handle only two distinct delay (default, custom1, custom2).
* All non-filtered packets must be set to band 1, other packets must be send to specified delays.
* We still want to provide some simple API.

This is starting to be a challenge :-)

1. Adding a prio qdisc with custom number of bands (10 in this example), all packets are send to band 0, handle id is `1:`
```
tc qdisc add dev eth0 root handle 1: prio bands 10 priomap 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
```

2. Adding a netem child to set delay, this child is added to band `1` (`1:2` means qdisc `1` band `2-1=1`), this child has handleid `30`
```
tc qdisc add dev eth0 parent 1:2 handle 30: netem delay 150ms
```

3. Add filter that will send specific packet to band 1 `priority 2 ... flowid 1:2` (all other packets are in band 0 = without delay)
```
tc filter add dev eth0 protocol ip parent 1: priority 2 u32 match ip dst 172.16.100.10 flowid 1:2
```

Repeating steps 2 and 3 with can define different delays for different bands.

pbrezina: This however starts to be over-complicated, we would need to add individual methods like `qdisc_add_prio`, `qdisc_add_netem`, `add_filter` and so on.

I think the most common use case is to set a delay to all outgoing connections, so it would be enough to stick to the current version (with some modification - you still need to set all unfiltered packets to one band) and say you can call it only once (define the limitations). Providing a replace/delete method if one needs to change the delay. WDYT?

pbrezina: More generic API can always be created later, or we can add a method to just run tc command with user given arguments given the command complexity. We can just take care of the teardown and obtaining list of interfaces.

pbrezina (commented):
Hi, I think this is good. But I don't think it will work correctly on more complex setup when you will want to have more delays and different delays.

pytest_mh/utils/tc.py
pbrezina: Actually, the maximum amount of bands is 16 so we can hardcode this. Then we can support up to 15 delays.

* create prio queue (if not exist) with 16 bands, send everything to 0
* add a delay and assign a band to it, remember the band
* add a filter and assign it to band based on requested delay

Some delete API should also be handy to remove particular filter.

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def setup_when_used(self):
+        """
+        Setup traffic control configuration.
+
+        :meta private:
+        """
+        super().setup_when_used()
+
+        # Let's find out the availabel interfaces
```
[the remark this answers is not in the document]
ikerexxe: Fixed

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,74 @@
...
+        ip = self.host.ssh.run(f"dig +short {hostname}", log_level=SSHLog.Error)
+
+        for interface in self.__interfaces:
+            self.host.ssh.run(
+                f"""
+                set -e
+                tc qdisc add dev {interface} root handle 1: prio
+                tc qdisc add dev {interface} parent 1:3 handle 30: netem delay {time}
+                tc filter add dev {interface} protocol ip parent 1:0 priority 3 u32 match ip dst {ip.stdout} flowid 1:3
+            """
```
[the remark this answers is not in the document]
ikerexxe: Not applicable anymore

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,74 @@
...
+    def add_delay(self, hostname: str, time: int):
+        """
+        Add delay to network connection.
+
+        :param hostname: Target hostname.
+        :type hostname: str
+        :param time: Delay in milliseconds.
+        :type time: int
+        """
+        ip = self.host.ssh.run(f"dig +short {hostname}", log_level=SSHLog.Error)
```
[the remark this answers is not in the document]
ikerexxe: Fixed

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,74 @@
...
+        self.host.ssh.run(tear, log_level=SSHLog.Error)
+        super().teardown_when_used()
+
+    def add_delay(self, hostname: str, time: int):
+        """
+        Add delay to network connection.
+
+        :param hostname: Target hostname.
+        :type hostname: str
+        :param time: Delay in milliseconds.
```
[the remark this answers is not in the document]
ikerexxe: Fixed

pytest_mh/utils/tc.py
ikerexxe: I think I understood what you mean in the previous comments, but let me explain what I did.

`add_delay()` keeps the same interface as before, but currently I set a limitation to the number of connections that can be delayed at a time to 15. Band 0 is used for the default and non-delayed connections. Band 1 to 15 are used for delayed traffic, and each hostname that adds a delay uses a band. It is recommended to set the delays from minimum to maximum to avoid starvation. All this information is provided in the API comments.

Another option that came to my mind and that can increase the number of delays in parallel, is to divide the bands per delay range. Thus, a delay of 1-100ms could be assigned to band 1, a delay of 101-200ms band 2, etc. I don't mind implementing that as it would give us more flexibility and the user doesn't have to be aware of any limitation. In any case, do you think we'll be using more than 1 or 2 delays at a time?

Regarding the API to remove a particular filter, let's agree on the delay addition first. It already took me some time to figure this out and I think I'll have to change it following your feedback.

Finally, please next time try to avoid having several contradictory comments without a final statement of what is actually expected (or isn't). It was difficult to follow the discussion and I had to read it several times.

ikerexxe:
Please close all resolved conversations.

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,93 @@
...
+            commands += (
+                f"tc qdisc add dev {interface} parent 1:{self.__band} handle {self.__band*10}: netem "
+                f"delay {time_unit}\n"
+            )
+            for ip in ip_list:
+                commands += (
+                    f"tc filter add dev {interface} protocol ip parent 1:0 u32 match ip dst {ip} "
+                    f"flowid 1:{self.__band}\n"
+                )
+            self.__restore.add(f"tc qdisc del dev {interface} root")
```
pbrezina, line 91: This should be moved directly to teardown - you call "add root" in setup for all interfaces, so you need to delete it all in teardown.

pbrezina (changes requested):
Thank you. I think this already works, see two comments inside.

It would be also good to provide `remove_delay` method that would remove the filter. But we can do it in other pull request when needed, if you don't want to spend any more time on it.

pytest_mh/utils/tc.py
```diff
@@ -0,0 +1,74 @@
...
+
+        :meta private:
+        """
+        tear: str = ""
+        for iter in self.__restore:
+            tear += iter + "\n"
+        self.host.ssh.run(tear, log_level=SSHLog.Error)
+        super().teardown_when_used()
+
+    def add_delay(self, hostname: str, time: int):
```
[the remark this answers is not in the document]
ikerexxe: One question for the last case, should I use the same band for all ip addresses in the list?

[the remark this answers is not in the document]
pbrezina: Since the delay is the same for all addresses, then yes.

But actually, you this was a leftover comment from some previous review session. I think this change would make it more generic, but I don't think it is really needed and I am happy with latest interface. Feel free to disregard it.

pytest_mh/utils/tc.py
ikerexxe: Done

ikerexxe:
I added the `remove_delay()` interface in a new commit

pytest_mh/utils/tc.py
pbrezina: This does not apply anymore. But the fix is `interface: str | None = None`

ikerexxe: Thanks for clarifying!

PR #23, opened 2023-11-06
fs: journalctl added

patriki01:


pbrezina:
Hi, do not add it to fs utils but convert it into standalone `JournaldUtils` class. Besides this, we will wait for Dan to provide the list of common tasks that we would like to implement as individual methods.

pytest_mh/utils/journal.py
```diff
@@ -0,0 +1,77 @@
...
+from .. import MultihostHost, MultihostUtility
+from ..ssh import SSHProcessResult
+
+__all__ = ["JournaldUtils"]
+
+
+class JournaldUtils(MultihostUtility):
+    """
+    Perform journald related tasks.
+
```
pbrezina, line 14: Remove blank line.

pytest_mh/utils/journal.py
```diff
@@ -0,0 +1,77 @@
...
+        unit: str | None = None,
+        lines: int | None = None,
+        since: str | None = None,
+        reverse: bool = False,
+        no_pager: bool = False,
+        grep: str | None = None,
+        output: str | None = None,
+        identifier: str | None = None,
+        system: bool = False,
+        user: bool = False,
```
pbrezina, line 36: Add `args` to support additional arguments if needed.

pbrezina (changes requested):
Hi, it may be better to name the file as `journald.py` since everyone knows it as journald.

It looks good. Now please collaborate with Dan to get the use cases.

danlavu:
Looking at our existing test cases. 

1. Get the log messages from **only** the duration of the test execution, currently we are clearing logs before running the tests. For example,  we will do this before a  login attempt; clear, login attempt, search logs, clear, 2nd login attempt, search logs, clear...  It doesn't have to be a clear, we can use timestamps or a diff the log messages. 
3. Search the logs for a string(s), regex , return True if found
4. Search the logs for a string(s), regex, return match count

patriki01:
I added those test cases. But I do not know if this kind of clearing systemctl with timestamp is sufficient. Please take a look and if necessary propose improvement. @pbrezina @danlavu

pytest_mh/_private/multihost.py
```diff
@@ -412,7 +412,7 @@ def setup(self) -> None:
         """
         Called before execution of each test.
         """
-        pass
+        self.test_start = self.ssh.exec(["date", "--rfc-3339=ns"]).stdout[0:19]
```
pbrezina, line 415: This should go to JournaldUtils.setup

pbrezina (changes requested):

pytest_mh/utils/journald.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        Called before execution of each test.
+        """
+        self.clear()
+
+    def clear(self) -> None:
+        """
+        Reset timestamp
+        """
+        self.test_start = self.host.ssh.exec(["date", "--rfc-3339=ns"]).stdout[0:19]
```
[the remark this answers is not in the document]
pbrezina: If you want to clear it, then you should use `--vacuum` 

But if this is the intention, then I'm fine with it to unblock your work. But I'd like to rework it, if possible, to "archive existing journal", "vacuum between tests", "restore old journal" to properly clean up after test run and instead of relying on timestamp. But this does not have to be part of this PR.

pytest_mh/utils/journald.py
```diff
@@ -0,0 +1,124 @@
...
+        lines: int | None = None,
+        since: str | None = None,
+        reverse: bool = False,
+        no_pager: bool = False,
+        grep: str | None = None,
+        output: str | None = None,
+        identifier: str | None = None,
+        system: bool = False,
+        user: bool = False,
+        args: list[str] | None = None,
```
pbrezina, line 50: list[Any] here as well :-)

pbrezina (changes requested):

pytest_mh/utils/journald.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        Called before execution of each test.
+        """
+        self.clear()
+
+    def clear(self) -> None:
+        """
+        Reset timestamp
+        """
+        self._test_start = self.host.ssh.exec(["date", "--rfc-3339=ns"]).stdout[0:19]
```
pbrezina, line 34: Please, add log_level=SSHLog.Error

pytest_mh/utils/journald.py
patriki01: Oh, of course

danlavu:
@patriki01 I won't be sure if it'll be sufficient until I use it.  ;)

pytest_mh/utils/journald.py
```diff
@@ -0,0 +1,126 @@
...
+            "grep": (cli.option.VALUE, grep),
+            "output": (cli.option.VALUE, output),
+            "identifier": (cli.option.VALUE, identifier),
+            "system": (cli.option.SWITCH, system),
+            "user": (cli.option.SWITCH, user),
+        }
+
+        return self.host.ssh.exec(["journalctl"] + cli.args(builder) + args, raise_on_error=False)
+
+    def is_match(self, pattern: str) -> bool:
```
danlavu, line 105: I think we should just call this match, the *is*, is implied.

PR #24, opened 2023-11-09
requirements: avoid installing ssh2-python

pbrezina:
ssh2-python is a dependency of parallel-ssh but does not currently
work on python-3.12. We use ssh-python in parallel-ssh so we can
avoid installing this dependency until it is fixed.


pbrezina:
It does not work. We need to wait for https://github.com/pypa/pip/pull/10837.

PR #30, opened 2023-11-22
artifacts: add support for dynamic artifacts list

pbrezina:
Now roles, utils or individual tests can add custom artifacts to 
role.artifacts. This will be added to the list of host artifacts configured
in mhc.yaml.


pbrezina:
I plan to use this feature to automatically collect core dumps if generated, but it is generically useful also if you need to collect some special logs in various tests (like IPA logs from server mode tests).

justin-stephenson (approved):
Ack, LGTM.

PR #32, opened 2023-12-04
utils: reword firewall module

pbrezina:
This commit adds outbound rules and changes the API so it is not
so much tied to single usecase. Now it is more flexible and extensible.

---

This breaks current API but so far SSSD and other IdM project that
we control are the only users so we can still afford it.


pytest_mh/utils/firewall.py
```diff
@@ -88,8 +82,145 @@ def parse_port_spec(self, spec: PortSpec) -> tuple[int, ProtocolSpec]:
...
+
+            firewall.inbound.drop_port(389)  # missing protocol defaults to "tcp"
+            firewall.inbound.drop_port((389, "tcp"))
+            firewall.inbound.drop_port([389, 636])
+
+        :param port: Port (integer), (port, protocol) tuple.
+        :type port: PortSpec | list[PortSpec]
+        """
+        pass
+
```
justin-stephenson, line 141: No `reject_port`?

pytest_mh/utils/firewall.py
```diff
@@ -88,8 +82,145 @@ def parse_port_spec(self, spec: PortSpec) -> tuple[int, ProtocolSpec]:
...
+    @abstractmethod
+    def drop_host(self, host: HostSpec | list[HostSpec]) -> None:
+        """
+        Drop outgoing traffic to given host.
+
+        The hostname is resolved to IPv4 and IPv6 addresses via DNS and
+        an accept rule is created for each available address.
```
justin-stephenson, line 208: incorrect copy/paste

pytest_mh/utils/firewall.py
pbrezina: No, Windows Firewall does not support reject, only drop.

danlavu: I can't think of a QE scenario in which we distinguish the difference between dropping and rejecting traffic, just as long as it's dropped, which both do. I think one function is better.

pytest_mh/utils/firewall.py
```diff
@@ -115,101 +257,415 @@ def teardown_when_used(self):
...
+
+    def reject_port(self, port_or_service: ServiceSpec | list[ServiceSpec]) -> None:
+        """
+        Reject incoming traffic on given port or service.
+
+        .. code-block:: python
+
+            firewall.inbound.reject_port(389)  # missing protocol defaults to "tcp"
+            firewall.inbound.reject_port((389, "tcp"))
+            firewall.inbound.reject_port("ldap")
```
danlavu, line 461: This is a lot of repetition. Maybe, three examples for the first function in the class, then any additional function one example?

pytest_mh/utils/firewall.py
```diff
@@ -115,101 +257,415 @@ def teardown_when_used(self):
...
+        """
+        self.__add_port(port_or_service, action="reject")
+
+    def drop_port(self, port_or_service: ServiceSpec | list[ServiceSpec]) -> None:  # type: ignore[override]
+        """
+        Drop incoming traffic on given port or service.
+
+        .. code-block:: python
+
+            firewall.inbound.drop_port(389)  # missing protocol defaults to "tcp"
```
danlavu, line 475: See previous comment.

pytest_mh/utils/firewall.py
```diff
@@ -115,101 +257,415 @@ def teardown_when_used(self):
...
+
+        self.firewall: Firewalld = firewall
+
+    def accept_port(self, port_or_service: ServiceSpec | list[ServiceSpec]) -> None:  # type: ignore[override]
+        """
+        Accept outgoing traffic on given port or service.
+
+        .. code-block:: python
+
+            firewall.outbound.accept_port(389)  # missing protocol defaults to "tcp"
```
danlavu, line 527: Changing my mind a little bit, I honestly think it's enough to have the one set of examples in the first function, the rest I think it's unnecessary.

danlavu:
Besides the redundant comments, this looks great.

pytest_mh/utils/firewall.py
pbrezina: Which function do you want to have documented?

Keep in mind, that IDE shows you docstring only of the current function that you are typing. If the docstring is present only in single place, it makes sense for sphinx doc, but you will loose that information in IDE.

pytest_mh/utils/firewall.py
pbrezina: reject = return immediately, useful for testing sssd in offline mode and you don't have to wait for timeout
drop = you must wait for timeout to bring sssd offline, useful when testing timeouts

pytest_mh/utils/firewall.py
pbrezina: Actually, it seems that this is no longer the case and help() shows the inherited docstring. I updated the patch so the documentation is only in the base class.

pbrezina:
Thank you, please, see updated patch.

pytest_mh/utils/firewall.py
jakub-vavra-cz: We should have reject here, in the API as an abstract method. Windows fw will ether not implement it or just use drop in implementation. Rejecting traffic is useful to have test faster (reject = return immediately).

pbrezina: It is not needed to have reject in abstract class. You will have it if needed on linux providers, just not on AD. It is also inbound traffic, that you are not going to use anyway. Outbound rules will be called from client where reject is available.

The exception is raised automatically if method is not provided via @abstractmethod.

pytest_mh/utils/firewall.py
danlavu: Thank you.

pytest_mh/utils/firewall.py
danlavu: That's good to know, and it makes sense. Because rejecting the server responds with access denied,  and dropping won't respond at all. So the only one we'd need is drop.

justin-stephenson (approved):
Ack, thank you.

danlavu (approved):

pytest_mh/utils/firewall.py
pbrezina: To clarify further, reject is not firewall thing. You can do it in iptables and we can use it to avoid waiting for timeouts during testing, but it is actually for routers. Implementing Window's reject and drop would mess up expectations about the test and since we can't implement it everywhere, it should not be part of generic API.

But really, what we want to use mostly are the outbound rules (as was requested on the mailing list and also in some pull request) on the linux client which will provide all three: accept, drop and reject. So calling `client.firewall.outbound.reject_host(provider)` will work for all providers as expected.

jakub-vavra-cz (approved):
Based on discussion I am giving this a green light.

pbrezina:
https://github.com/SSSD/sssd-test-framework/pull/63 awaits Dan's approval, therefore I have unpushed this PR in order to not break existing code. I'll push it once it is approved.

PR #33, opened 2023-12-22
Fs: diff utility added

patriki01:
Propose additional arguments if you can think of any.


pytest_mh/utils/fs.py
```diff
@@ -608,6 +608,45 @@ def wc(
 
         return self.host.ssh.exec(["wc", *args, file], log_level=SSHLog.Error)
 
+    def diff(
+        self,
+        file1: str,
+        file2: str,
```
pbrezina, line 614: path1, path2 since you also allow recursive which operates on folders

pbrezina:
I took the liberty and fixed it myself so I can do a new release.

pbrezina (approved):

PR #34, opened 2024-01-11
SYSTEMD: Reset the failed flag in all SystemdServices::*start() methods

aplopez:
Calls `systemctl reset-failed $SERVICE` on all `*start()` methods.

This is required by changes to sssd-test-framework in https://github.com/SSSD/sssd-test-framework/pull/69

### DIFF
--- a/pytest_mh/utils/services.py
+++ b/pytest_mh/utils/services.py
@@ -46,7 +46,8 @@ class SystemdServices(MultihostUtility):
         self.__set_initial_state(service)
         self.logger.info(f'systemd: starting "{service}" asynchronously')
         return self.host.ssh.async_run(
-            f'systemctl start "{service}" || systemctl status "{service}"', log_level=SSHLog.Error
+            f'systemctl reset-failed "{service}"; systemctl start "{service}" || systemctl status "{service}"',
+            log_level=SSHLog.Error,
         )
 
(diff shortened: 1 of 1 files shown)


pytest_mh/utils/services.py
```diff
@@ -46,7 +46,8 @@ def async_start(self, service: str) -> SSHProcess:
         self.__set_initial_state(service)
         self.logger.info(f'systemd: starting "{service}" asynchronously')
         return self.host.ssh.async_run(
-            f'systemctl start "{service}" || systemctl status "{service}"', log_level=SSHLog.Error
+            f'systemctl reset-failed "{service}" && systemctl start "{service}" || systemctl status "{service}"',
```
pbrezina, line 49: Use `;` instead of `&&`. `reset-failed` may fail in the unit is stopped and not enabled by default (this is triggered by autofs tests in sssd). Using `;` is fine, if the service fails to start and `reset-failed` failed, we will see it in the logs. And if by any chance `reset-failed` failed but start succeeded, we don't care.

Same for restart.

aplopez: Agreed.

pbrezina:
Please, run black on the code to fix the ci failure. `pip3 install black` and `black .`

PR #36, opened 2024-01-16
artifacts: extract archive by default

pbrezina:
Artifacts are now extracted by default which is more convenient for
most use cases. For local run, you want to be able to quickly debug
failures and automation usually already compress everything in single
archive so it is again better to make it more easily accessible.

--mh-compress-artifacts was added to switch to the original behavior.


justin-stephenson (approved):
Testing works as expected with default extracting and also testing `--mh-compress-artifacts` option. Ack.

ikerexxe (approved):
I like it! :smile:

Issue #26, opened 2023-11-10
Find alternative to parallel-ssh


pbrezina:
parallel-ssh currently does not install on python-3.12 and unfortunately the project's maintainers are unresponsive and it seems that the project is dead. We will need to find a replacement, since we do not have a capacity to properly maintain our own fork.

See: https://github.com/ParallelSSH/ssh2-python/pull/194

https://github.com/ansible/pylibssh may work for us.

pbrezina:
Work in progress, I needed to add some libssh calls to pylibssh in order to allow async ssh I/O.

pylibssh: https://github.com/ansible/pylibssh/pull/576
pytest-mh changes: https://github.com/pbrezina/pytest-mh/tree/pylibssh

PR #44, opened 2024-02-09
ssh: switch from parallel-ssh to ansible-pylibssh

pbrezina:
Unfortunately, parallel-ssh has become a dead project incompatible with
python-3.12+. Some effort was created by its community to make it
compatible, unfortunately the only maintainer with write access is
unresponsive and there is not much hope for a new release.

Therefore we switch to ansible-pylibssh, which provides low level libssh
python bindings. There is significant performance impact in artificial test
case ("running 'exit 0' over and over again is twice as slow) due to
differences in poll implementation, however the impact is neglitable in
real world scenarios (~20 seconds delay on 350 test cases). This small
slowdown also provides a huge benefit in very small CPU usage 0-1% where
parallel-ssh keeps spinning on 10-11%.

Resolves: https://github.com/next-actions/pytest-mh/issues/26

### DIFF
--- a/pytest_mh/ssh.py
+++ b/pytest_mh/ssh.py
@@ -6,12 +6,12 @@ import shlex
 import textwrap
 from enum import Enum, auto
 from typing import Any, Generator, Type
+from collections.abc import Generator
 
 import colorama as c
-import pssh.clients.base.single
-import pssh.clients.ssh
-import pssh.exceptions
-import pssh.output
+from pylibsshext.errors import LibsshSessionException
+from pylibsshext.session import Session as LibsshSession
+from pylibsshext.channel import Channel as LibsshChannel
 
 from ._private.logging import MultihostLogger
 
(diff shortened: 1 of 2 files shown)


pbrezina:
This requires a changes to pylibssh that are not yet released, we are waiting for ansible-pylibssh 1.2.0.

* https://github.com/ansible/pylibssh/pull/576
* https://github.com/ansible/pylibssh/pull/578

pbrezina:
@justin-stephenson pylibssh 1.2.0 was finally released so this is ready for review.

The CI/Example fails dues to missing sssd-kcm package in the containers, Madhuri is going to fix it.

justin-stephenson (approved):
Ack, works for me.

Some minor comments below which still mention `pssh`

~~~
justin@fedora:~/github/pytest-mh$ grep -irn 'pssh'
pytest_mh/ssh.py:515:        pssh simply calls the command as $shell '$command', e.g.
pytest_mh/ssh.py:610:        pssh simply calls the command as $shell '$command', e.g.
~~~

pbrezina:
Good catch. Fixed.

PR #45, opened 2024-02-12
SYSTEMD: make `get_property()` with older 'systemctl'

alexey-tikhonov:
On centos-8 systemctl doesn't support '-P'
'--value --property' should work with both older and modern versions.


madhuriupadhye (approved):
LGTM

jakub-vavra-cz (approved):

PR #48, opened 2024-02-16
mh: support extra artifacts in host and utility as well

pbrezina:
For example, if we have SSSDUtils, we can automatically make it
to collect SSSD logs and database instead of relying on configuration
in mhc.yaml.

Additionally, it is possible to use prepare_artifacts to collect coredumps
from coredumpclt/journald and other stuff.


justin-stephenson (approved):
Ack, thank you.

PR #50, opened 2024-03-13
Gather artifacts from host and topology setup and teardown and reorganize logs

pbrezina:
This is a bigger patchset that adds the ability to get artifacts from
pytest_setup, pytest_teardown and topology_setup and topology_teardown.

It also add the ability to set dynymic logs for this operation as well
as for the test run itself.

It refactors the code quite a bit and reorginizes logs into a directory
tree so we are able to store all the requested artifacts.


pbrezina:
Merging this, since nobody has time to review patch set of this size.

PR #53, opened 2024-04-03
multihost: avoid using inspect.getmembers in MultihostUtility

pbrezina:
Using inspect.getmembers calls getattr internally which evalutes
lazy properties as a side effect. This is undesirable and can
cause troubles in some cases.

This patch switches to metaclass magic. It also moves the decorators
outside MultihostUtility class for simpler use.

### DIFF
--- a/pytest_mh/__init__.py
+++ b/pytest_mh/__init__.py
@@ -15,6 +15,8 @@ from ._private.multihost import (
     MultihostOSFamily,
     MultihostRole,
     MultihostUtility,
+    mh_utility_ignore_use,
+    mh_utility_used,
 )
 from ._private.plugin import MultihostPlugin, pytest_addoption, pytest_configure
 from ._private.topology import Topology, TopologyDomain
@@ -35,6 +37,8 @@ __all__ = [
     "MultihostRole",
     "MultihostTopologyControllerArtifacts",
     "MultihostUtility",
+    "mh_utility_ignore_use",
+    "mh_utility_used",
     "pytest_addoption",
     "pytest_configure",
     "Topology",
--- a/pytest_mh/_private/multihost.py
+++ b/pytest_mh/_private/multihost.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import inspect
-from abc import ABC, abstractmethod
+from abc import ABC, ABCMeta, abstractmethod
 from functools import wraps
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Generic, Type, TypeVar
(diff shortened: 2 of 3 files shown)


pbrezina:
This is currently causing issues in SSSD tests when connection to LDAP server is established sooner then expected because lazy property `conn` is evaulated by inspect.getmembers in MultihostUtility constructor. This PR removes it.

There is also other inspect.getmembers call which does not cause issues (since it is called later), but it will be removed as well but following PR. WIP: https://github.com/pbrezina/pytest-mh/tree/utils

justin-stephenson:
What would be a simple way to test `setup_when_used` and `teardown_when_used` ?

Also, would it make sense for default pytest-mh behavior to have `setup/teardown_when_used = true` ? What is the use case for `setup` over `setup_when_used` ??

danlavu:
> What would be a simple way to test `setup_when_used` and `teardown_when_used` ?
> 
> Also, would it make sense for default pytest-mh behavior to have `setup/teardown_when_used = true` ? What is the use case for `setup` over `setup_when_used` ??

I use those methods with the authselect tests, with-pamaccess and with-faillock features so the setup and teardown only happens when pam is modified.

danlavu (approved):
Looks good to me.

jakub-vavra-cz:
Just tested the change in the jenkins and it works as expected.

justin-stephenson:
> > What would be a simple way to test `setup_when_used` and `teardown_when_used` ?
> 
> You can create custom utility class and add it to the client role.
> 
> ```python
> def ExampleUtility(MultihostUtility):
>     def setup_when_used(self):
>         self.logger.info("I AM HERE SETUP")
> 
>     def teardown_when_used(self):
>         self.logger.info("I AM HERE TEARDOWN")
> ```
> 
> You should see the messages in `--mh-log-path=/dev/stderr` output
> 
> > Also, would it make sense for default pytest-mh behavior to have `setup/teardown_when_used = true` ? What is the use case for `setup` over `setup_when_used` ??
> 
> setup happens everytime, prior test execution. setup_when_used can be used to not execute setup for stuff that is not used often (e.g. sshd mentioned by Dan). It is also used in traffic control (don't mess up network settings if you don't need it).
> 
> However, I simplify it a lot in subsequent patches -> there [will be](https://github.com/pbrezina/pytest-mh/commit/fe91d58f4410559795bdea802fbd9ba231bb0cb3) just setup/teardown methods with the option to postpone execution setup for its first used. I also add a [contextmanager](https://github.com/pbrezina/pytest-mh/commit/4d48859e78cfc8991947877dd2a7e9933b1eed95) that can create utilities on demand inside test, so not all of them have to be in roles.

Thanks for clarification.

PR #54, opened 2024-04-10
Add support for reentrant utilities to MultihostHost and other improvements

pbrezina:
* avoid using getmembers
  * getmembers invokes lazy properties which is not desired
* rework setup and teardown code to be more straightforward
  * it is still pretty complicated, I'll try to improve it more later
* add mh_utility context managers
  * utils can be created adhoc within a test, so not every utility has to be part of a role class
* rework and simplify setup_when_used and teardown_when_used
  * these methods are removed and replaced with setup() and teardown()
  * it adds decorator to force postponed setup on utility
* add MultihostReentrantUtility
  * this adds context support to utility - it can be entered and exited multiple times to create different scopes
  * this may be useful in tests but it is especially required to enable utils in host objects with per-host and per-topology scopes
* setup and teardown utils in dependency order
  * if A requires B then A is always setup before B and B is teardown before A
* hosts now support automatic setup/enter/exit/teardown of reentrant utils
  * non reentrant utils can be used in hosts too but then host object is responsible for calling setup/teardown when needed/expected

Docs update is not sufficient. Given the amount of features that we added lately, I plan to rework the documentation to be more useful and complete in another PR.


pbrezina:
@danlavu @justin-stephenson we can have a call about this pr if you'd like some clarification

But essentially, it is important for us to be able to use utils to create configurations/join realm/create users/etc at host/topology setup. If you e.g. need to create 1000 users in topology setup and then remove it in topology teardown for the performance tests, currently you would need to do it by code duplication. With reentrant utils, you can move the code to util class and reuse it.

danlavu:
Should this work yet? 

```
test setup failed
request = <SubRequest 'mh' for <Function test_gpo__is_set_to_enforcing_and_users_are_allowed[su] (ad)>>

    @pytest.fixture(scope="function")
    def mh(request: pytest.FixtureRequest) -> Generator[MultihostFixture, None, None]:
        """
        Pytest multihost fixture. Returns instance of :class:`MultihostFixture`.
        When a pytest test is finished, this fixture takes care of tearing down the
        :class:`MultihostFixture` object automatically in order to clean up after
        the test run.
    
        .. note::
    
            It is preferred that the test case does not use this fixture directly
            but rather access the hosts through dynamically created role fixtures
            that are defined in ``@pytest.mark.topology``.
    
        :param request: Pytest's ``request`` fixture.
        :type request: pytest.FixtureRequest
        :raises ValueError: If not multihost configuration was given.
        :yield: MultihostFixture
        """
    
        data: MultihostItemData | None = MultihostItemData.GetData(request.node)
        if data is None:
            nodeid = f"{request.node.parent.nodeid}::{request.node.originalname}"
            raise ValueError(f"{nodeid}: mh fixture requested but no multihost configuration was provided")
    
        if data.multihost is None:
            raise ValueError("data.multihost must not be None")
    
        if data.topology_mark is None:
            raise ValueError("data.topology_mark must not be None")
    
        mh = MultihostFixture(request, data, data.multihost, data.topology_mark)
        try:
>           mh._enter()

../../../../pytest-mh/pytest_mh/_private/fixtures.py:455: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../../pytest-mh/pytest_mh/_private/fixtures.py:393: in _enter
    self._invoke_phase("SETUP ROLES", self._setup_roles)
../../../../pytest-mh/pytest_mh/_private/fixtures.py:364: in _invoke_phase
    cb()
../../../../pytest-mh/pytest_mh/_private/fixtures.py:264: in _setup_roles
    item.setup()
../../../../sssd-test-framework/sssd_test_framework/roles/client.py:98: in setup
    self.sssd.import_domain(domain, role)
../../../../pytest-mh/pytest_mh/_private/multihost.py:926: in wrapper
    return method(self, *args, **kwargs)
../../../../sssd-test-framework/sssd_test_framework/utils/sssd.py:349: in import_domain
    self.config["sssd"].setdefault("domains", "")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <configparser.ConfigParser object at 0x7fcf545ea890>, key = 'sssd'

    def __getitem__(self, key):
        if key != self.default_section and not self.has_section(key):
>           raise KeyError(key)
E           KeyError: 'sssd'

/usr/lib64/python3.11/configparser.py:979: KeyError
```

pbrezina:
Did you apply https://github.com/SSSD/sssd-test-framework/pull/99? It should solve this.

danlavu (approved):
This looks good to me.

justin-stephenson (approved):
Ack, thank you for your patience.

PR #56, opened 2024-05-28
journald: dump and collect journal when test ends

pbrezina:
This generates /var/log/journald.log when a test ends and marks it for collection.


danlavu (approved):
LG2M

PR #59, opened 2024-06-13
ssh: keep reading chunks if it is not possible to decode it

pbrezina:


justin-stephenson (approved):
Ack.

alexey-tikhonov:
@pbrezina I think patch is good but I'm a bit confused with naming.

`SSHOutputBuffer::_read()` is only used in `SSHOutputBuffer::send()`, right?

It's a bit confusing that `send()` actually "reads".
And why does it have "value" argument?

FWIW: a test run - https://github.com/SSSD/sssd/actions/runs/9502942770/job/26192967768?pr=7444

pbrezina:
send() is interface of Python's `Generator` class that we use (you may be familiar with `yield`).  You can read more about it e.g. [here](https://realpython.com/introduction-to-python-generators/).

alexey-tikhonov:
Thanks.

PR #60, opened 2024-06-18
plugin: add pytest_report_teststatus to MultihostUtility

pbrezina:
Utility classes can alter the test result using this hook.

PR #61, opened 2024-06-19
utils: add Auditd

pbrezina:
This utility will collect auditd logs for each test. It can be also 
configured to report any AVC denial and optionally fail the test if AVC
denial is found.

### DIFF
--- a/pytest_mh/utils/auditd.py
+++ b/pytest_mh/utils/auditd.py
@@ -1,0 +1,140 @@
+from __future__ import annotations
+
+import re
+from typing import Literal
+
+import pytest
+
+from .. import MultihostHost, MultihostUtility
+from ..ssh import SSHLog
+
+__all__ = ["Auditd"]
+
+
+class Auditd(MultihostUtility[MultihostHost]):
+    """
+    Auditd utilities.
+
+    Collects audit logs and detects AVC denials.
+    """
+
+    def __init__(
+        self,
+        host: MultihostHost,
+        *,
+        avc_mode: Literal["fail", "warn", "ignore"],
+        avc_filter: str | None = None,
+    ) -> None:
+        """
+        ``avc_mode`` values:
+
+        * ``ignore``: all failures are ignored
+        * ``warn``: test result category is set to "AVC DENIALS" and the test is
+          marked as such in a test summary, however test outcome and pytest exit
+          code is kept intact
+        * ``fail``: test result category is set to "AVC DENIALS" and the test is
+          marked as such in a test summary, if a test outcome is ``passed`` it
+          is set to ``failed`` and pytest will return non-zero exit code
+
+        :param host: Multihost host.
+        :type host: MultihostHost
+        :param avc_mode: Action taken when AVC denial is found in audit logs.
+        :type avc_mode: Literal["fail", "warn", "ignore"]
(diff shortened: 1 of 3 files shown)


pbrezina:
Depends on https://github.com/next-actions/pytest-mh/pull/60.

This requires virtual machine since auditd is not working in containers.

pytest_mh/utils/auditd.py
```diff
@@ -0,0 +1,133 @@
...
+            match = re.search(self.avc_filter, records)
+            if match is None:
+                return None
+
+            error_msg = (
+                f"AVC denial was detected with regex filter '{self.avc_filter}'. Check audit log of {report.nodeid}."
+            )
+
+        match self.avc_mode:
+            case "fail":
```
jakub-vavra-cz, line 126: It would make more sense to set the test to error status instead.
We would like to distinguish here different states:
1. test itself passed but AVC happened - ERROR or WARNING. (AVC happened but functionality works as expected)
2. test did not pass and avc is present - FAIL. (This could be a fatal avc blocking part of functionality test testing )
We might want to ignore avcs altogether if there is a know issue and we want to deliver results.

jakub-vavra-cz (changes requested):
I would like to see handling as an error.

pytest_mh/utils/auditd.py
```diff
@@ -0,0 +1,133 @@
...
+        if not records:
+            return None
+
+        error_msg = f"AVC denial was detected. Check audit log of {report.nodeid}."
+        if self.avc_filter:
+            match = re.search(self.avc_filter, records)
+            if match is None:
+                return None
+
+            error_msg = (
```
danlavu, line 121: It'd be nice to prefix the report with a 'sestatus' and 'getsebool -a | \ on', capturing enabled booleans and policy information.

pytest_mh/utils/auditd.py
pbrezina: Thanks, this makes sense. I change the semantics to keep track of test outcome.

```

Selected tests will use the following hosts:
  audit: audit.test

collected 545 items / 544 deselected / 1 selected                                                                                                                                                                                            

tests/test_audit.py::test_audit (audit) PASSED/AVC DENIAL

======= 544 deselected, 1 AVC DENIALS in 0.94s =======
```

FAILED/PASSED is kept in the verbose output and all tests with AVC errors are counted towards "AVC DENIALS" at the end of the test. If "fail" mode is set, then non-zero return code is returned from pytest. If "warn" is set then test result is respected - so if test pass, zero exit code will be returned.

pytest_mh/utils/auditd.py
pbrezina: I removed the message since it is not showed anywhere in the latest version.

Anyway, ausearch is able to track any AVC denials, that means also AppArmor. We could add this output elsewhere, maybe preferably to downstream CI as individual step (like gather all package versions, selinux settings, ...)?

danlavu: That's fine, it not going to be looked at often and will probably be more of a nuisance  to see it all the time. ;)

justin-stephenson (approved):
LGTM, Ack.

jakub-vavra-cz (approved):

PR #62, opened 2024-06-24
fs: add sed support

aborah-sudo:
Add sed functionality to new test framework

### DIFF
--- a/pytest_mh/utils/fs.py
+++ b/pytest_mh/utils/fs.py
@@ -738,3 +738,22 @@ class LinuxFileSystem(MultihostReentrantUtility):
         mode = f"{user}" if user else ""
         mode += f":{group}" if group else ""
         return self.host.ssh.exec(["chown", mode, *path.split(), *args], log_level=SSHLog.Error)
+
+    def sed(self, command: str, path: str, args: list[str] | None = None) -> SSHProcessResult:
+        """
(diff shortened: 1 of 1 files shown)


pytest_mh/utils/fs.py
```diff
@@ -738,3 +738,22 @@ def chown(
...
+        :type command: str
+        :param path: File where changes will happen
+        :type path: str
+        :param args: Additional options, defaults to None
+        :type args: list[str] | None, optional
+        :return: Result of process
+        :rtype: SSHProcessResult
+        """
+        self.backup(path)
+        self.logger.info(f'Changing file "{path}"')
```
pbrezina, line 757: Hi, can you log full sed command here? That would be more helpful in the logs. E.g.:

`Running sed s/a/b/g on /etc/path`

aborah-sudo: Done

jakub-vavra-cz:
@aborah-sudo Looks like You need to run the code through black to fix the ci.

aborah-sudo:
> @aborah-sudo Looks like You need to run the code through black to fix the ci.

Done

aborah-sudo:
> Hi, thank you. It looks good.
> 
> Before we push it, can you just rename the commit subject to `fs: add sed support` to follow the convention of other commits?

Done

pbrezina (approved):

PR #64, opened 2024-06-28
utils: add Coredumpd

pbrezina:
Collects generated core files from /var/lib/systemd/coredump and if there
are any, optionally fail the test.

### DIFF
--- a/pytest_mh/utils/coredumpd.py
+++ b/pytest_mh/utils/coredumpd.py
@@ -1,0 +1,205 @@
+from __future__ import annotations
+
+import re
+from typing import Literal
+
+import pytest
+
+from .. import MultihostArtifactsType, MultihostHost, MultihostUtility
+from ..ssh import SSHLog
+from .fs import LinuxFileSystem
+
+__all__ = ["Coredumpd"]
+
+
+class Coredumpd(MultihostUtility[MultihostHost]):
+    """
+    Coredumpd utilities.
+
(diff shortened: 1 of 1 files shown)


pbrezina:
You will need VM to test this as coredumps produced from with in containers are still stored on the host. There is ongoing work to enable to store the coredump directly in the container, so we'll likely get that in the future as well.

https://github.com/systemd/systemd/pull/29272
https://github.com/containers/podman/issues/22958

pytest_mh/utils/coredumpd.py
```diff
@@ -0,0 +1,205 @@
...
+        # Ignore if no core files matches the filter
+        if self.filter:
+            match = re.search(self.filter, "\n".join(self._corefiles))
+            if match is None:
+                return None
+
+        original_outcome = report.outcome
+
+        # Fail the test if fail mode is selected
+        if report.outcome == "passed" and self.mode == "fail":
```
justin-stephenson, line 200: what is the reason to have `if report.outcome == "passed"` here?

pbrezina: If the test passed, we want to switch it to fail, otherwise don't touch the outcome. In theory, the outcome here may only be  "passed" or "failed" so this check is redundant. But I added it for verbosity and future safety.

justin-stephenson (approved):
Ack thank you Pavel.

danlavu (approved):
Looks good.

Thank you, Pavel.

PR #66, opened 2024-07-02
errors: switch to ExceptionGroup

pbrezina:
python-3.11 brings ExceptionGroup feature which is very useful for
our use case where we need to teardown multiple objects eventhough
some of them failed.

With ExceptionGroup, no exception and its traceback is lost.

See:
https://peps.python.org/pep-0654/


ikerexxe (approved):
LGTM!

PR #67, opened 2024-07-11
conn: add container/podman/docker connection

pbrezina:
This pull requests introduce a generic connection class so we
are no longer bounded to SSH. It also implements a container
connection so podman or docker can be used instead.

This brings some breaking changes:
* mhc.yaml: `ssh` configuration was replaced with `conn`
* `host.ssh` was replaced with `host.conn`
* generic classes should be used instead of SSH specific: SSHClient -> Connection, SSHProcess -> Process, SSHProcessResult -> ProcessResult and so on.

Example configuration for SSH:

```yaml
  - hostname: dc.ad.test
    role: ad
    os:
      family: windows
    conn:
      type: ssh
      username: Administrator@ad.test
      password: vagrant
    config:
      adminpw: vagrant
      client:
        ad_domain: ad.test
```

Example configuration for podman:

```yaml
  - hostname: master.ipa.test
    role: ipa
    conn:
      type: podman
      name: ipa
      sudo: True
    config:
      client:
        ipa_domain: ipa.test
        krb5_keytab: /enrollment/ipa.test.keytab
        ldap_krb5_keytab: /enrollment/ipa.test.keytab
```


pbrezina:
Documentation is not included. After this is merged, I plan to restructuralize and rewrite the documentation so it can contain all the new features we added lately.

jakub-vavra-cz:
Why not keep the "ssh:" section intact?
It would be better to add an additional configuration item "conn_type:" instead with default value ssh. That can be switched to podman or docker.

pbrezina:
> Why not keep the "ssh:" section intact? It would be better to add an additional configuration item "conn_type:" instead with default value ssh. That can be switched to podman or docker.

I want to keep the amount of reserved keywords minimal. This way, we have reserved only "conn", instead of "conn_type", "ssh", "podman" and "docker".

pytest_mh/conn/container.py
```diff
@@ -0,0 +1,459 @@
...
+
+        self.__popen = None
+        self.__stdout = None
+        self.__stderr = None
+        self.__stdin = None
+
+
+class ContainerConnectionError(ConnectionError):
+    """
+    Unable to authenticate over Container.
```
justin-stephenson, line 329: s/authenticate/connect/ ? Nothing related to authentication

justin-stephenson:
> 
> Example configuration for podman:
> 
> ```yaml
>   - hostname: master.ipa.test
>     role: ipa
>     conn:
>       type: podman
>       name: ipa
>    ...
> ```


Why not `container` instead of `name` here? When first looking at this it wasn't clear to me what `name` should be (arbitrary or otherwise)

pbrezina:
> > Example configuration for podman:
> > ```yaml
> >   - hostname: master.ipa.test
> >     role: ipa
> >     conn:
> >       type: podman
> >       name: ipa
> >    ...
> > ```
> 
> Why not `container` instead of `name` here? When first looking at this it wasn't clear to me what `name` should be (arbitrary or otherwise)

Great idea. Thank you. Done.

justin-stephenson (approved):
Ack, thank you.

danlavu (commented):
The code looks good! I haven't had a chance to test it yet. Do you want me to approve it, or can I get to it early next week?

thalman (approved):
Great work, thanks

PR #68, opened 2024-07-12
FIREWALL: Use nftables instead of direct rules

aplopez:
Direct rules use the `iptables` framework, which is no longer supported in RHEL 10.

To block outbound traffic, we start using the `nftables` framework instead, but we use it through the `nft` command because `firewall-cmd` doesn't support this. We use the tables and chains created by `firewalld` for a better integration. In particular, because those rules will be deleted when `firewalld` is restarted or its configuration reloaded, which is the behavior expected by the test framework.

### DIFF
--- a/pytest_mh/utils/firewall.py
+++ b/pytest_mh/utils/firewall.py
@@ -223,6 +223,15 @@ class Firewalld(Firewall):
     All changes are automatically reverted when a test is finished.
     """
 
+    NFT_TABLE_FAMILY: str = "inet"
+    """The IP family for the nft table"""
+
+    NFT_TABLE_NAME: str = "firewalld"
+    """The table for nft rules """
+
+    NFT_OUTPUT_CHAIN: str = "filter_OUTPUT"
+    """The output chain in the ``NFT_TABLE_NAME`` table"""
+
     def __init__(self, host: MultihostHost) -> None:
         """
         :param host: Remote host instance.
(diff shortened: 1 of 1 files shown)


pytest_mh/utils/firewall.py
```diff
@@ -289,81 +301,59 @@ def _next_priority(self) -> int:
...
-        priority: int | None = None,
-    ) -> int:
+    ) -> None:
         """
-        Add a new direct rule.
+        Add a new nft rule.
+
+        This method adds a nftables rule to the firewall. This rule is added without
+        the firewall knowing it but, if it is added to the firewall's table and chain,
+        it will be removed when the firewall is reloaded or restarted.
```
ikerexxe, line 314: The second sentence doesn't make much sense. Do you mean that if we add this rule to the firewall it'll be removed when the firewall is reloaded?

ikerexxe (commented):
The changes do not seem major, but I have some questions that I would like to clarify before continuing with the review.

pytest_mh/utils/firewall.py
aplopez: The seconds sentence means the firewall is not aware of this rule because it didn't add it itself (as it was done before with firewall-cmd's direct rules.

Because the rule is added to a table and chain created by the firewall, they will be removed when the firewall is removed. I guess it simple erases the table with all the rules it has in it.

pytest_mh/utils/firewall.py
```diff
@@ -223,6 +223,18 @@ class Firewalld(Firewall):
...
+    """The IP family for the nft table"""
+
+    NFT_TABLE_NAME: str = "firewalld"
+    """The table for nft rules """
+
+    NFT_OUTPUT_CHAIN: str = "filter_OUTPUT"
+    """The output chain in the ``NFT_TABLE_NAME`` table"""
+
+    NFT_INPUT_CHAIN: str = "filter_INPUT"
+    """The input chain in the ``NFT_TABLE_NAME`` table"""
```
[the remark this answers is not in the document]
aplopez: I'm not planing on doing it because it is not currently necessary. But one day we may need it.

pbrezina:
Hi, I am not going to actively review this, I will let Iker and Tomas handle it. But out of curiosity, can you talk to firewalld devs and learn what are their plans? Will they remove direct rules? Will they switch it to use nftables? The direct rules are mostly needed only if you want to set outbound rules, I know there has been some work in this area, but using direct rules was quite easier (and documented for this purpose in rhel docs) when I was working on this.

aplopez:
> can you talk to firewalld devs and learn what are their plans?

I just did. They gave me an alternative solution that works creating a permanent policy but creating dynamic (non-permanent) rich rules. What surprises me is that I did try this method at some point but I must have been doing something wrong because it didn't work for me at that time.

So will will modify this PR to use that method.

thalman (commented):
The patch looks good, and it is fine for replacing direct rules. Could/did you test it with RHEL8? The tables looks slightly different there but I think that it should work.

My general comment it that we should avoid using this in tests in favor of rich_rule which is more "platform independent". Of course the rich rule can't block outgoing traffic so it must be used on target host. This approach is also closer to real life scenario when connection is blocked somewhere else (hw firewall, target machine...)

pbrezina:
> The patch looks good, and it is fine for replacing direct rules. Could/did you test it with RHEL8? The tables looks slightly different there but I think that it should work.
> 
> My general comment it that we should avoid using this in tests in favor of rich_rule which is more "platform independent".

firewalld is using nftables to implement rich rules, so it is safe to use nftables directly - it will be available.

> Of course the rich rule can't block outgoing traffic so it must be used on target host. This approach is also closer to real life scenario when connection is blocked somewhere else (hw firewall, target machine...)

This was the original idea, but in the end we added outgoing traffic as well. I don't remember the rationale expect that Jakub wanted it, but we reached an agreement.

aplopez:
I created a separate PR with the alternative solution using rich rules and a policy. I'm not closing this one until we all agree to use the new one. Please switch to https://github.com/next-actions/pytest-mh/pull/69

aplopez:
> The patch looks good, and it is fine for replacing direct rules. Could/did you test it with RHEL8? The tables looks slightly different there but I think that it should work.

It was tested 2 days ago. Check this comment: https://github.com/next-actions/pytest-mh/pull/68#issuecomment-2230816607

> My general comment it that we should avoid using this in tests in favor of rich_rule which is more "platform independent". Of course the rich rule can't block outgoing traffic so it must be used on target host. This approach is also closer to real life scenario when connection is blocked somewhere else (hw firewall, target machine...)

So I guess you'll love the alternative solution in [PR#69](https://github.com/next-actions/pytest-mh/pull/69).

thalman:
> > The patch looks good, and it is fine for replacing direct rules. Could/did you test it with RHEL8? The tables looks slightly different there but I think that it should work.
> > My general comment it that we should avoid using this in tests in favor of rich_rule which is more "platform independent".
> 
> firewalld is using nftables to implement rich rules, so it is safe to use nftables directly - it will be available.
> 

That is correct, but my concern is that chains can be redesigned/renamed and our tests will stop working eventually.
Using rich rules looks more "firewalld" oriented and so it looks more stable solution for me. This my comment is not about this PR but more about how we should write the test.

Anyway if it is acceptable to have permanent change at the startup (empty policy created) then the PR#69 is the way forward without nft.

aplopez:
Closing this PR in favor of [PR#69](https://github.com/next-actions/pytest-mh/pull/69) with a better solution.

PR #69, opened 2024-07-17
FIREWALL: Use rich rules instead of direct rules

aplopez:
Direct rules have been deprecated for some time now, so we switch to using rich rules.

However, we need a specific policy to which the rule will be applied. This policy is permanent (a requirement) and is created at setup, deleted at teardown, and remains active during the whole test. Only the rules are dynamic.

This works on RHEL 8 thourgh 10.

### DIFF
--- a/pytest_mh/utils/firewall.py
+++ b/pytest_mh/utils/firewall.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from abc import abstractmethod
+from random import randrange
 from typing import Any, Literal, TypeAlias
 
 from .. import MultihostHost, MultihostRole, MultihostUtility
@@ -232,6 +233,7 @@ class Firewalld(Firewall):
 
         self.__inbound: FirewalldInboundRules = FirewalldInboundRules(self)
         self.__outbound: FirewalldOutboundRules = FirewalldOutboundRules(self)
+        self._policies: list[str] = []
 
         self._priority: int = 30000
         """
@@ -244,12 +246,24 @@ class Firewalld(Firewall):
         not remove "accept" rule but the "drop" rule takes precedence.
         """
 
+    def setup(self) -> None:
+        """
+        Set the firewall up.
+
+        :meta private:
+        """
+        super().setup()
+        self.add_policy(f"test-policy-{randrange(99999)}", ingress="HOST", egress="ANY")
+        self.host.conn.exec(["firewall-cmd", "--reload"], log_level=ProcessLogLevel.Error)
+
     def teardown(self) -> None:
         """
         Revert all firewall changes.
 
         :meta private:
         """
+        for policy in list(self._policies):
+            self.remove_policy(policy)
         self.host.conn.exec(["firewall-cmd", "--reload"])
         super().teardown()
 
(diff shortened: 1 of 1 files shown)


pytest_mh/utils/firewall.py
```diff
@@ -244,12 +245,17 @@ def __init__(self, host: MultihostHost) -> None:
         not remove "accept" rule but the "drop" rule takes precedence.
         """
 
+        self._policy: str = f"test-policy-{randrange(99999)}"
+        self.add_policy(self._policy, ingress="HOST", egress="ANY")
+        self.host.ssh.exec(["firewall-cmd", "--reload"], log_level=SSHLog.Error)
```
pbrezina, line 250: Please, move this to setup().

aplopez: I should have guessed that there was a `setup()` function if there was a `teardown()` one.
Done.

pytest_mh/utils/firewall.py
```diff
@@ -289,95 +302,77 @@ def _next_priority(self) -> int:
...
-        :type table: str, optional
-        :param ip_family: If the rules is added as IPv4, IPv6 rule or both, defaults to ``all``.
-        :type ip_family: Literal["ipv4", "ipv6", "all"], optional
-        :param priority: Rule priority, defaults to None (= auto-assign next value)
-        :type priority: int | None, optional
-        :return: Rule priority, to be used for rule removal.
+        Add a new (permanent) policy.
+
+        Except for the name, all parameters are optional. When the priority is not provided,
+        the next priority is assigned. When the other parameters is not provided,
```
ikerexxe, line 318: s/is/are

pytest_mh/utils/firewall.py
```diff
@@ -289,95 +302,77 @@ def _next_priority(self) -> int:
         self._priority -= 1
         return priority
 
-    def add_direct_rule(
+    def add_policy(
         self,
-        chain: str,
-        args: list[Any],
+        name: str,
```
ikerexxe, line 307: This parameter is missing in the description

pytest_mh/utils/firewall.py
```diff
@@ -289,95 +302,77 @@ def _next_priority(self) -> int:
...
         :rtype: int
         """
+        self.logger.info(f'Firewalld: adding policy "{name}"')
+
+        cmd = ["firewall-cmd", "--permanent", "--new-policy", name]
+        self.host.ssh.exec(cmd, log_level=SSHLog.Error)
+        cmd[2] = "--policy"
+
         if priority is None:
             priority = self._next_priority
```
ikerexxe, line 340: I wonder, shouldn't this variable change at some point?

pytest_mh/utils/firewall.py
```diff
@@ -289,95 +302,77 @@ def _next_priority(self) -> int:
...
-        priority: int,
-        chain: str,
-        args: list[Any],
-        *,
-        table: str = "filter",
-        ip_family: Literal["ipv4", "ipv6", "all"],
-    ) -> None:
-        """
-        Remove direct rule.
+    def remove_policy(self, name: str) -> None:
```
ikerexxe, line 354: I know it's a simple method but I'd expect some comment explaining what it does and the parameters.

pytest_mh/utils/firewall.py
```diff
@@ -541,15 +547,13 @@ def __add_port(
             else:
                 port, protocol = self.firewall.parse_port_spec(item)
 
-            self.firewall.add_direct_rule(
-                chain="OUTPUT", args=["-p", protocol, "--dport", port, "-j", action], table="filter"
-            )
+            self.firewall.add_rich_rule(f"port port={port} protocol={protocol} {action}")
```
ikerexxe, line 550: You wrote `port` two times. Is that correct?

ikerexxe (commented):
Some inline comments, apart from that it looks good.

pytest_mh/utils/firewall.py
aplopez: Fixed

pytest_mh/utils/firewall.py
aplopez: Fixed

pytest_mh/utils/firewall.py
aplopez: It's not a variable but a property. It changes every time it is used. Check line 290.

pytest_mh/utils/firewall.py
aplopez: Done

pytest_mh/utils/firewall.py
ikerexxe: Now I see the `@property` attribute. Thanks!

ikerexxe (approved):
LGTM. Thanks for all the explanations!

pytest_mh/utils/firewall.py
```diff
@@ -289,95 +302,85 @@ def _next_priority(self) -> int:
...
-        table: str = "filter",
-        ip_family: Literal["ipv4", "ipv6", "all"] = "all",
+        ingress: str | None = None,
+        egress: str | None = None,
         priority: int | None = None,
+        target: Literal["CONTINUE", "ACCEPT", "DROP", "REJECT"] | None = None,
     ) -> int:
         """
-        Add a new direct rule.
+        Add a new (permanent) policy.
```
pbrezina, line 315: But it is still removed in teardown, right?

aplopez: I see your point. It's done now.

pbrezina (commented):
I didn't review the code, but it looks ok conceptually.

As a rule of thumb - don't call any commands in __init__, use setup for that. This way, it is possible to call setup only when something from the utils is actually used (see mh_postpone_setup). And every change done must be reverted in teardown so we do not polute other tests.

pbrezina:
Bump @thalman

thalman (approved):
Sorry, I did the review earlier and forgot to ACK it. Approving now...

pbrezina:
@aplopez Please, resolve conflict and we can merge.

PR #81, opened 2024-09-13
topology-controller: make _init a public method

pbrezina:
This method is inteded to be overwritten, therefore it should be public.


justin-stephenson (approved):
Ack, thank you.

PR #82, opened 2024-09-13
topology-marker: allow to pass TopologyMark directly, make _CreateFromArgs public and docs improvements

pbrezina:
In addition to Topology and KnownTopologyBase values.

```
@pytest.mark.topology(TopologyMark(...))
@pytest.mark.topology([TopologyMark(...), ...])
```


justin-stephenson (approved):
Ack.

PR #83, opened 2024-09-13
topology-marker: make TopologyMark._CreateFromArgs public

pbrezina:
This method is intended to be overriden in subclasses therefore it
should be public.


pbrezina:
Superseded by https://github.com/next-actions/pytest-mh/pull/82

PR #84, opened 2024-09-13
fixtures: implement additional fixtures to get data from mh

pbrezina:
These are getters for selected MultihostFixture properties.


justin-stephenson (approved):
Ack.

PR #85, opened 2024-09-16
fs: Add restorecon to fs utility

jakub-vavra-cz:
The files/dirs created by the fs utility needs to have proper selinux context.

### DIFF
--- a/pytest_mh/utils/fs.py
+++ b/pytest_mh/utils/fs.py
@@ -85,6 +85,7 @@ class LinuxFileSystem(MultihostReentrantUtility):
             """,
             log_level=ProcessLogLevel.Error,
         )
+        self.restorecon(path)
 
     def mkdir_p(
         self, path: str, *, mode: str | None = None, user: str | None = None, group: str | None = None
@@ -113,6 +114,7 @@ class LinuxFileSystem(MultihostReentrantUtility):
             """,
             log_level=ProcessLogLevel.Error,
         )
+        self.restorecon(path)
 
         if result.stdout and result.stdout != path:
             if not backup_exists:
@@ -276,6 +278,7 @@ class LinuxFileSystem(MultihostReentrantUtility):
             input=contents,
             log_level=ProcessLogLevel.Error,
         )
+        self.restorecon(path)
 
     def append(
         self,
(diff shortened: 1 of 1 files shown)


pytest_mh/utils/fs.py
```diff
@@ -757,3 +763,23 @@ def sed(self, command: str, path: str, args: list[str] | None = None) -> Process
         self.logger.info(f"Running sed {command} on {path}")
         args = args if args else []
         return self.host.conn.exec(["sed", *args, command, path], log_level=ProcessLogLevel.Error)
+
+    def restorecon(self, path: str) -> bool:
+        """
+        Restore selinux context on a file or directory.
+        Does nothing when restorecon is not present on os without selinux.
```
ikerexxe, line 770: s/on/in

pytest_mh/utils/fs.py
```diff
@@ -757,3 +763,23 @@ def sed(self, command: str, path: str, args: list[str] | None = None) -> Process
...
+
+        :param path: File or directory where changes will happen
+        :type path: str
+        :return: True if restorecon succeeded
+        :rtype: bool
```
ikerexxe, line 775: I am concerned that this function will return `False` if `restorecon` is not present in the system, as this may be seen as an error by the caller. I wonder if this shouldn't be an `int/enum` type, which will help us detect different types of scenarios.

pytest_mh/utils/fs.py
jakub-vavra-cz: done

pbrezina:
I wouldn't merge this until we understand what happens and confirm that this will actually fix it. The problem was with /etc/krb5.conf, which we don't modify using the fs util, and it is present only in FIPS mode. I would like to avoid binding this class to SELinux, unless it serves some use case.

pytest_mh/utils/fs.py
jakub-vavra-cz: The idea is that this will be noop on distros without selinux.
If we need to make sure that restorecon exists the fs utility can already do that before calling this.
We do not need to complicate this one.

ikerexxe: Current implementation isn't a noop as any caller checking the return value for this method where SELinux isn't present would get a `False` value and this could be interpreted as a failure. If we are to return a boolean, then the non-SELinux case should return `True`

pbrezina:
IIRC you said that everything works fine in FIPS mode yesterday, then this is not in fact needed. Please, reopen it if I understood it wrongly.

PR #86, opened 2024-09-26
journald: set no-pager parameter

pbrezina:
Running journalctl in pager (interactive) mode does not make any sense
in automated tests. On the other hand, we want to sent this parameter
by default to make sure it is never run in interactive mode.


jakub-vavra-cz (approved):
LGTM.

PR #88, opened 2024-10-22
add unit tests and fix some trivial issues

pbrezina:
I started writing extensive unit tests, this is the first batch. I found some very minor issues that are fixed here as well.


pbrezina:
@justin-stephenson May I ask you for a review? I don't think it is necessary to deep dive into every single test case.

justin-stephenson:
> @justin-stephenson May I ask you for a review? I don't think it is necessary to deep dive into every single test case.

Sure, how do I run the tests locally?

pbrezina:
> > @justin-stephenson May I ask you for a review? I don't think it is necessary to deep dive into every single test case.
> 
> Sure, how do I run the tests locally?

Either run pytest directly or via tox.

tests/test_conn.py
```diff
@@ -0,0 +1,161 @@
...
+    assert shell.shell_command == "powershell -NonInteractive -Command"
+
+    cmd = shell.build_command_line("echo hello world", cwd="/home/test", env={"HELLO": "WORLD", "JOHN": "DOE"})
+    expected = textwrap.dedent(
+        """
+        $Env:HELLO = WORLD
+        $Env:JOHN = DOE
+        cd /home/test
+
+        echo hello world
```
justin-stephenson, line 158: better to use powershell commands like `Set-Location` and `Write-Output` here?

justin-stephenson (approved):
Tests LGTM, I didn't review in detail but ack.

tests/test_conn.py
pbrezina: It does not matter since the code is not really executed, but cd and echo are valid commands. I think they are actually aliases.

justin-stephenson: Ok no problem.

PR #98, opened 2025-01-29
multihost: add conn to role and utility as a shortcut

pbrezina:
The main purpose of pytest-mh is to allow executing commands on the
hosts. Therefore it is natural to provide conneciton via a shortcut
instead of writing `host.conn` everywhere.


pbrezina:
Closing. It's better to leave this on the user since there might be different kinds of connections required by the tests (ldap for example). The user can choose what the shortcut should be (ldap or ssh connection).

PR #99, opened 2025-02-17
 markers: new preferred_topology marker

danlavu:
Reducing test running time and system resources, we can now mark
 a test with @pytest.mark.preferred_topology(KnownTopology.topology),
 making the topology the single and default run unless the marker is
 overridden. This is intended to be used for tests that do not need
 to cover all topologies all the time.

### DIFF
--- a/docs/articles/tips-and-tricks/preferred-topology.rst
+++ b/docs/articles/tips-and-tricks/preferred-topology.rst
@@ -1,0 +1,34 @@
+Preferred Topology
+==================
+
+The preferred topology marker ``pytest.mark.preferred_topology`` can be
+used if you want the flexibility to run against all topologies, but do not
+need to do it all the time. An example could be Nightly builds or a CI where
+a single topology is adequate coverage. This will reduce the resources
+required to run the tests and their execution times. This is possible by using
+the ``pytest.mark.preferred_topology`` marker.
+
+This marks the test with a default topology, deselecting any additional
+topologies. ``--mh-ignore-preferred-topology`` can be used to ignore the
+marker.
+
+If more than one preferred topology has been defined, only the last topology
+will be used. If the preferred topology contains no value, the marker is
+ignored.The ``pytest.mark.preferred_topology`` marker accepts three types of
+values, :class:`~pytest_mh.TopologyMark` value of
+:class:`~pytest_mh.KnownTopologyBase` or ``str``.
+
+.. code-block:: python
+    :caption:  Example: set preferred topology to IPA
+
+
+    @pytest.mark.topology(KnownTopologyGroup.AnyProvider)
+    @pytest.mark.preferred_topology(KnownTopology.IPA)
+    def test_preferred_mark_known_topology_ipa():
+        pass
+
+
+    @pytest.mark.topology(KnownTopologyGroup.AnyProvider)
+    @pytest.mark.preferred_topology("ipa")
+    def test_preferred_mark_string_ipa():
+        pass
--- a/pytest_mh/_private/plugin.py
+++ b/pytest_mh/_private/plugin.py
@@ -18,7 +18,7 @@ from .data import MultihostItemData
 from .errors import TeardownExceptionGroup
 from .fixtures import MultihostFixture
 from .logging import MultihostLogger
-from .marks import TopologyMark
+from .marks import KnownTopologyBase, TopologyMark
 from .multihost import (
     MultihostArtifactsMode,
     MultihostConfig,
@@ -63,6 +63,7 @@ class MultihostPlugin(object):
         self.mh_collect_artifacts: MultihostArtifactsMode = pytest_config.getoption("mh_collect_artifacts")
         self.mh_artifacts_dir: Path = Path(pytest_config.getoption("mh_artifacts_dir"))
         self.mh_compress_artifacts: bool = pytest_config.getoption("mh_compress_artifacts")
+        self.mh_ignore_preferred_topology: bool = pytest_config.getoption("mh_ignore_preferred_topology")
 
         # Read --mh-collect-logs, default to --mh-collect-artifacts
         self.mh_collect_logs: MultihostArtifactsMode = pytest_config.getoption("mh_collect_logs")
@@ -200,6 +201,7 @@ class MultihostPlugin(object):
         self.logger.info(f"  collect artifacts: {self.mh_collect_artifacts}")
         self.logger.info(f"  artifacts directory: {self.mh_artifacts_dir}")
         self.logger.info(f"  collect logs: {self.mh_collect_logs}")
+        self.logger.info(f"  ignore-preferred-topology: {self.mh_ignore_preferred_topology}")
         self.logger.info("")
 
         signal(SIGINT, self.sigint_handler)
@@ -521,6 +523,32 @@ class MultihostPlugin(object):
     def _is_multihost_required(self, item: pytest.Item) -> bool:
         return item.get_closest_marker(name="topology") is not None
 
+    def _can_run_preferred_topology(self, mark: pytest.Mark, current_topology: str, item: pytest.Item) -> bool:
+        if len(mark.args) != 1:
+            raise ValueError(
+                f"{item.nodeid}: Unexpected number of arguments to pytest.mark.preferred_topology: "
+                f"got {len(mark.args)}, expected 1"
+            )
+
+        arg = mark.args[0]
+
+        if isinstance(arg, KnownTopologyBase):
+            name = arg.value.name
+        elif isinstance(arg, TopologyMark):
+            name = arg.name
+        elif isinstance(arg, str):
+            name = arg
+        else:
+            raise ValueError(
+                f"{item.nodeid}: Unexpected type of pytest.mark.preferred_topology: "
+                f"got {type(arg)}, expected KnownTopologyBase | TopologyMark | str"
+            )
+
+        if name == current_topology:
+            return True
+
+        return False
+
     def _can_run_test(self, item: pytest.Item, data: MultihostItemData | None) -> bool:
         if data is None:
             return not self._is_multihost_required(item)
@@ -546,6 +574,12 @@ class MultihostPlugin(object):
 
             if data.topology_mark.name not in self.mh_topology:
                 return False
+
+        # Run only for preferred topology unless specific topology is requested or the marker is ignored
+        if not self.mh_topology and not self.mh_ignore_preferred_topology:
+            preferred_topology = item.get_closest_marker(name="preferred_topology")
+            if preferred_topology is not None and data.topology_mark is not None:
+                return self._can_run_preferred_topology(preferred_topology, data.topology_mark.name, item)
 
         return True
 
@@ -793,6 +827,12 @@ def pytest_addoption(parser):
     parser.addoption("--mh-lazy-ssh", action="store_true", help="Postpone connecting to host SSH until it is required")
 
     parser.addoption(
+        "--mh-ignore-preferred-topology",
+        action="store_true",
+        help="All topologies will run, ignore the preferred_topology marker",
+    )
+
+    parser.addoption(
         "--mh-topology",
         action="append",
         default=[],
@@ -860,6 +900,13 @@ def pytest_configure(config: pytest.Config):
         "markers",
         "require(condition, reason): evaluate condition, parameters may be topology fixture, "
         "the test is skipped if condition is not met",
+    )
+
+    config.addinivalue_line(
+        "markers",
+        "preferred_topology(topology: KnownTopologyBase | TopologyMark | str): "
+        "mark test with a preferred topology."
+        "Test will execute once, skipping additional topologies",
     )
 
     config.pluginmanager.register(MultihostPlugin(config), "MultihostPlugin")
--- a/docs/articles/tips-and-tricks.rst
+++ b/docs/articles/tips-and-tricks.rst
@@ -11,3 +11,4 @@ inspiration for your project.
     tips-and-tricks/building-command-line
     tips-and-tricks/features-detection
     tips-and-tricks/pytest-fixtures
+    tips-and-tricks/preferred-topology
--- a/pytest_mh/_private/data.py
+++ b/pytest_mh/_private/data.py
@@ -54,7 +54,7 @@ class MultihostItemData(object):
 
     @staticmethod
     def GetData(item: pytest.Item) -> MultihostItemData | None:
-        return item.stash[DataStashKey]
+        return item.stash.get(DataStashKey, None)
 
 
 DataStashKey = pytest.StashKey[MultihostItemData | None]()


danlavu:
I have it working, but I don't understand this attribute error from tox, still looking into it. 

```
lint: commands[3]> mypy --install-types --non-interactive pytest_mh example tests docs
Found 2 errors in 1 file (checked 50 source files)
pytest_mh/_private/plugin.py:401: error: Item "str" of "Mark | str" has no attribute "args"  [union-attr]
pytest_mh/_private/plugin.py:402: error: Item "str" of "Mark | str" has no attribute "args"  [union-attr]
```

I don't like the logging output; it prints 'SKIPPED' and 'Skipped' for one test case; I'm not sure where it's coming from. 

```
PASSED [ 31%]SKIPPED [ 37%]Single run marker set, skipping ipa

Skipped
```

The following marker values are tested, logs attached.

```
@pytest.mark.importance("critical")
@pytest.mark.topology(KnownTopologyGroup.AnyProvider)
def test_identity__single_run_no_marker(client: Client, provider: GenericProvider):


@pytest.mark.single_run()
@pytest.mark.importance("critical")
@pytest.mark.topology(KnownTopologyGroup.AnyProvider)
def test_identity__single_run_no_value(client: Client, provider: GenericProvider):


@pytest.mark.single_run("ad")
@pytest.mark.importance("critical")
@pytest.mark.topology(KnownTopologyGroup.AnyProvider)
def test_identity__single_run_value_specified(client: Client, provider: GenericProvider):


@pytest.mark.single_run("ad")
@pytest.mark.single_run("samba")
@pytest.mark.importance("critical")
@pytest.mark.topology(KnownTopologyGroup.AnyProvider)
def test_identity__single_run_invalid_values_specified(client: Client, provider: GenericProvider):
```

[pytest-single-run.txt](https://github.com/user-attachments/files/18878209/pytest-single-run.txt)

pytest_mh/_private/plugin.py
```diff
@@ -384,6 +386,31 @@ def pytest_runtest_setup(self, item: pytest.Item) -> None:
...
+
+        for i in item.iter_markers("single_run"):
+            if len(i.args) > 0:
+                self.single_run_topology = i.args[0]
+            else:
+                for y in item.iter_markers("topology"):
+                    if len(y.args) > 0 and y.args[0].name == "AnyProvider":
+                        self.single_run_topology = "ldap"
+                    if len(y.args) > 0 and y.args[0].name == "AnyADProvider":
+                        self.single_run_topology = "samba"
```
pbrezina, line 408: This is SSSD specific, can't be here. Please, remove it.

pytest_mh/_private/plugin.py
```diff
@@ -384,6 +386,31 @@ def pytest_runtest_setup(self, item: pytest.Item) -> None:
...
+
+        if self.single_run_topology is not self.current_topology:
+            self.logger.info(f"Single run marker set, skipping {self.current_topology}")
+            pytest.skip()
```
pbrezina, line 412: Move this into a standalone method.

pbrezina (changes requested):
I'm not a big fun of `single_run`. How about `preferred_topology`? Its still not perfect, so suggestion is welcomed. You should also not take the string but the topology marker object instead. It would be also good to allow lists. That is:

```python
# By default, runs only LDAP
@pytest.mark.preferred_topology(KnownTopology.LDAP)
@pytest.mark.topology(KnownTopologyGroup.AnyProvider)
def test_exampe(...):
    pass

# By default, runs only LDAP and IPA
@pytest.mark.preferred_topology(KnownTopology.LDAP)
@pytest.mark.preferred_topology(KnownTopology.IPA)
@pytest.mark.topology(KnownTopologyGroup.AnyProvider)
def test_exampe(...):
    pass

# By default, runs only LDAP and IPA
@pytest.mark.preferred_topology([KnownTopology.LDAP, KnownTopology.IPA])
@pytest.mark.topology(KnownTopologyGroup.AnyProvider)
def test_exampe(...):
    pass
```

danlavu:
> I'm not a big fun of `single_run`. How about `preferred_topology`? Its still not perfect, so suggestion is welcomed. You should also not take the string but the topology marker object instead. It would be also good to allow lists. That is:
> 
 
Linguistically, `preferred_topology` is unclear. It sounds like it makes the defined topology run first, while in reality, the new default is that it only runs against the selected topology. Is it the phrase or one of the words in particular? Maybe the negation sounds better, @mark.skip_extra_topologies? I want it to be clear that it skips/bypasses/ignores/omits additional topologies and is an exclusive/unique/single/solo/distinctive run. One of those words might make it sound better?

danlavu:
"preferred topology" is fine. Fewer characters are better, just as long as the documentation is good and clear, which I'll do.

docs/articles/life-cycle/skipping-tests.rst
```diff
@@ -166,3 +166,35 @@ tests and are not available at this point.
...
+
+    .. grid-item-card::  Examples of @pytest.mark.preferred_topology()
+
+        .. tab-set::
+
+            .. tab-item:: @pytest.mark.preferred_topology()
+
+                .. code-block:: python
+
+                        @pytest.mark.preferred_topology("ldap")
```
pbrezina, line 188: Don't work with names, we need to work mainly with objects here. Strings can be accepted as well to support the ad hoc topologies, but we want to use `KnownHost.LDAP` here as the main use case.

docs/articles/life-cycle/skipping-tests.rst
```diff
@@ -166,3 +166,35 @@ tests and are not available at this point.
                     tests/test_passkey.py::test_skip__lambda (ldap) SKIPPED (Server is not built with password policy support)
                     tests/test_passkey.py::test_skip__function (ldap) SKIPPED (Server is not built with password policy support)
                     tests/test_passkey.py::test_skip__function_and_reason (ldap) SKIPPED (Server is not built with password policy support)
+
+Preferred topology
+===================
+
+Sometimes, you want to maintain the flexibility to run tests against all topologies but may
+not need to run all topologies all the time. It is possible to run the test once, skipping any
+additional topologies using ``pytest.mark.preferred_topology``.
```
pbrezina, line 175: Little bit more description of the use case would be nice.

docs/articles/life-cycle/skipping-tests.rst
```diff
@@ -166,3 +166,35 @@ tests and are not available at this point.
...
+
+        .. tab-set::
+
+            .. tab-item:: @pytest.mark.preferred_topology()
+
+                .. code-block:: python
+
+                        @pytest.mark.preferred_topology("ldap")
+                        @pytest.mark.topology(KnownTopologyGroup.AnyProvider)
+                        def test_single_run__no_value(client: Client, provider: GenericProvider):
```
pbrezina, line 190: add "pass" or "assert True" to make it a valid python code

pytest_mh/_private/plugin.py
```diff
@@ -562,6 +566,25 @@ def _clone_function(self, name: str, f: pytest.Function) -> pytest.Function:
             originalname=f.originalname,
         )
 
+    def _preferred_topology(self, item: pytest.Item):
+        if self.mh_ignore_preferred_topology:
+            return
+
+        if len([i for i in item.iter_markers("preferred_topology")]) > 1:
+            raise ValueError("Too many arguments for @pytest.mark.preferred_topology.")
```
pbrezina, line 574: Hmm, how about taking the mark that is the closest?

Topologies can be inherited from class, this should be possible to inherit as well. I.e.

```python
@pytest.mark.preferred_topology(KnownTopology.LDAP)
@pytest.mark.topology(KnownTopologyGroup.AnyProvider)
class TestFoo():
   def test_bar():
        pass

    @pytest.mark.preferred_topology(KnownTopology.IPA)
    def test_foo():
        pass
```

The preferred topology for test_bar will be LDAP, for test_foo IPA.

pytest_mh/_private/plugin.py
```diff
@@ -862,6 +891,12 @@ def pytest_configure(config: pytest.Config):
         "the test is skipped if condition is not met",
     )
 
+    config.addinivalue_line(
+        "markers",
+        "preferred_topology(topology: str | None): mark test with a preferred topology."
```
pbrezina, line 896: I would like to see:

`str | TopologyMark | list[str | TopologyMark]` and support these to make it more flexible

pbrezina (changes requested):
Hi, this looks good, but I think it it too restrictive. Let's make it more flexible.

pytest_mh/_private/plugin.py
danlavu: We only expect a single value, so why accept more than one by adding lists?

danlavu:
@pbrezina, 

@andreboscatto was brainstorming this idea, which sounded like ``` @pytest.mark.not_this_topology()```.  Which led me to show him the pytest-mh code. He pointed out some flaws in my logic, which should be an improvement.  Another question, when writing tests, we should avoid things like, 

```
    if isinstance(provider, (IPA)):
        pytest.skip()
```
vs. 

```
@pytest.mark.topology(KnownTopology.LDAP)
@pytest.mark.topology(KnownTopology.AD)
@pytest.mark.topology(KnownTopology.Samba)
```

The cost of this is way less, correct?

docs/articles/life-cycle/skipping-tests.rst
```diff
@@ -166,3 +166,37 @@ tests and are not available at this point.
                     tests/test_passkey.py::test_skip__lambda (ldap) SKIPPED (Server is not built with password policy support)
                     tests/test_passkey.py::test_skip__function (ldap) SKIPPED (Server is not built with password policy support)
                     tests/test_passkey.py::test_skip__function_and_reason (ldap) SKIPPED (Server is not built with password policy support)
+
+Preferred topology
+===================
+
+Sometimes, you want to maintain the flexibility to run tests against all topologies, but not always. Nightly builds and CI are examples where running a test against a single topology by default may make more sense than all topologies. This can reduce the resources required to run the tests and their execution times. All while using the same test code. This is possible by using the ``pytest.mark.preferred_topology`` marker. This marks the test with a default topology that will run, and for the comprehensive runs, "--mh-ignore-preferred-topology" will ignore this marker entirely.
```
pbrezina, line 173: Keep 80 chars limit for text.

pbrezina: It might be better to move this to tips and tricks section.

docs/articles/life-cycle/skipping-tests.rst
```diff
@@ -166,3 +166,37 @@ tests and are not available at this point.
...
+
+
+.. note::
+    If the CLI argument '--mh-topology' is used, the test will use that value and ignore the 'preferred_topology' marker.
```
pbrezina, line 177: I think it would be better to provide more thorough description of the behavior it the main text instead of relying on a side note.

pytest_mh/_private/plugin.py
```diff
@@ -63,6 +65,7 @@ def __init__(self, pytest_config: pytest.Config) -> None:
         self.mh_collect_artifacts: MultihostArtifactsMode = pytest_config.getoption("mh_collect_artifacts")
         self.mh_artifacts_dir: Path = Path(pytest_config.getoption("mh_artifacts_dir"))
         self.mh_compress_artifacts: bool = pytest_config.getoption("mh_compress_artifacts")
+        self.mh_ignore_preferred_topology: bool = pytest_config.getoption("mh_ignore_preferred_topology")
```
pbrezina, line 68: print value in https://github.com/next-actions/pytest-mh/blob/master/pytest_mh/_private/plugin.py#L202

pytest_mh/_private/plugin.py
```diff
@@ -562,6 +567,22 @@ def _clone_function(self, name: str, f: pytest.Function) -> pytest.Function:
...
+        if self.mh_ignore_preferred_topology:
+            self.logger.info(" --mh-ignore-preferred-topology set: Ignoring preferred topology marker!")
+            return
+
+        if len(self.mh_topology) > 0:
+            self.logger.info(" --mh-topology set: Ignoring preferred topology marker!")
+            return
+
+        for i in item.iter_markers("preferred_topology"):
+            self.preferred_topology = i.args[0] if isinstance(i.args[0], str) else i.args[0].value.name
```
pbrezina, line 580: This can be local variable only.

pytest_mh/_private/plugin.py
```diff
@@ -562,6 +567,22 @@ def _clone_function(self, name: str, f: pytest.Function) -> pytest.Function:
...
+
+            self.logger.info(f" {item.name}: Preferred topology ({self.preferred_topology})")
```
pbrezina, line 582: the message should go to pytest.skip(msg)

docs/articles/life-cycle/skipping-tests.rst
pbrezina: It is also important to document the semantics - what happens if multiple markers are used and so on, what are the possible values and so on.

pytest_mh/_private/plugin.py
```diff
@@ -562,6 +567,22 @@ def _clone_function(self, name: str, f: pytest.Function) -> pytest.Function:
...
         )
 
+    def _preferred_topology(self, item: pytest.Item):
+        if self.mh_ignore_preferred_topology:
+            self.logger.info(" --mh-ignore-preferred-topology set: Ignoring preferred topology marker!")
+            return
+
+        if len(self.mh_topology) > 0:
+            self.logger.info(" --mh-topology set: Ignoring preferred topology marker!")
+            return
```
pbrezina, line 577: This is logged for each test. Once, during pytest initialization, is more than enough. At https://github.com/next-actions/pytest-mh/blob/master/pytest_mh/_private/plugin.py#L202 to make it visible.

pytest_mh/_private/plugin.py
```diff
@@ -562,6 +567,22 @@ def _clone_function(self, name: str, f: pytest.Function) -> pytest.Function:
...
+
+        for i in item.iter_markers("preferred_topology"):
```
pbrezina, line 579: Since the semantics is "pick closest" instead of "merge", it will be better to use `item.get_closest_marker`

pytest_mh/_private/plugin.py
pbrezina: Also be more thorough when checking the type to make sure only allowed types were passed.

pbrezina (changes requested):
I would like to see list of topologies as possible value as well, as a list or *args.

It works, unless you set more than one marker. But this will be fixed by using get_closest_marker instead of iter_markers. There are some issues, easy to fix though, see my comments inside.

However, I wonder, what is your reason to skip the test instead of deselecting it (i.e. do not attempt to run it at all)? There is downside for it, we attempt to run the test which means setup and teardown code is executed even if the test is skipped. We could probably find a better place for it, where we would skip the setup and teardown, but we would still be executing host and topology setup and teardown even if these hosts and topologies are all filtered out by preferred_topology.

Deselecting the test instead of skipping it would avoid this. But I admit, it may be just a corner case.

docs/articles/life-cycle/skipping-tests.rst
danlavu: I moved it to tips and tricks. I was planning on dropping it entirely. If speed and fewer resources are the goal, this **that** should be the path. The choice to skip was your recommendation, so we knew it what was being skipped, do you want me to drop it earlier, during collection?

danlavu:
@pbrezina This is ready for another look. 

A couple of major differences: it now deselects the test in modifyItems. Added --mh-preferred-topology, to override the marker for ad hoc runs to select a different topology.

docs/articles/tips-and-tricks/preferred-topology.rst
```diff
@@ -0,0 +1,69 @@
+Preferred topology
+===================
+
+If you want the flexibility to run tests against all topologies, but do not
+have the need to do it all the time. An example could be Nightly builds
+or and CI where a single topology is adequate coverage. This will reduce
```
pbrezina, line 6: or a CI

docs/articles/tips-and-tricks/preferred-topology.rst
```diff
@@ -0,0 +1,69 @@
+Preferred topology
+===================
+
+If you want the flexibility to run tests against all topologies, but do not
```
pbrezina, line 4: The preffered topology marker (``pytest.mark.preferred_topology``) can be used if you ...

pytest_mh/_private/plugin.py
```diff
@@ -63,6 +64,8 @@ def __init__(self, pytest_config: pytest.Config) -> None:
         self.mh_collect_artifacts: MultihostArtifactsMode = pytest_config.getoption("mh_collect_artifacts")
         self.mh_artifacts_dir: Path = Path(pytest_config.getoption("mh_artifacts_dir"))
         self.mh_compress_artifacts: bool = pytest_config.getoption("mh_compress_artifacts")
+        self.mh_ignore_preferred_topology: bool = pytest_config.getoption("mh_ignore_preferred_topology")
+        self.mh_preferred_topology: str = pytest_config.getoption("mh_preferred_topology")
```
pbrezina, line 68: What is the use case?

As I understand it, it does not fit well into the main use case, which is to run test only on certain topology by default. We want to decide, what the preferred topology for each test is. Two tests may have different preferred topologies, therefore overriding this setting for all tests to run against one specific topology does not really make sense to me. I would just remove it, with this understanding.

pytest_mh/_private/plugin.py
```diff
@@ -288,6 +293,13 @@ def pytest_collection_modifyitems(self, config: pytest.Config, items: list[pytes
                 deselected.append(item)
                 continue
 
+            # If mh_ignore_preferred_topology is True. Skip the preferred topology entirely.
+            if not self.mh_ignore_preferred_topology:
+                if item.get_closest_marker(name="preferred_topology") is not None:
+                    if not self._run_preferred_topology(item, data):
```
pbrezina, line 299: Pass the marker as argument instead of fetching it again in the called method.

pytest_mh/_private/plugin.py
```diff
@@ -288,6 +293,13 @@ def pytest_collection_modifyitems(self, config: pytest.Config, items: list[pytes
...
+                        deselected.append(item)
```
pbrezina, line 300: This call should be moved to `_can_run_test` where the decision for other cases is made.

pytest_mh/_private/plugin.py
```diff
@@ -549,6 +561,38 @@ def _can_run_test(self, item: pytest.Item, data: MultihostItemData | None) -> bo
...
+        if self.mh_preferred_topology is not None:
+            preferred_topology = self.mh_preferred_topology.lower()
+        else:
+            if preferred_mark is not None:
+                if preferred_mark.args[0] is not None:
+                    # Preferred mark accepts TopologyMark which is enumEnum or a string
+                    if isinstance(preferred_mark.args[0], enum.Enum):
+                        preferred_topology = preferred_mark.args[0].name.lower()
+                    else:
+                        preferred_topology = preferred_mark.args[0].lower()
```
pbrezina, line 584: This needs to be more bullet proof.

```python

        if len(preferred_mark.args) != 0:
            raise ValueError(...)

        arg = preferred_mark.args[0]
        if isinstance(preferred_mark, KnownTopologyBase):
            ...
        elif isinstance(preferred_mark, TopologyMark):
            ...
        elif isinstance(preferred_mark, str):
            ...
        else:
            raise ValueError(...)
```

pytest_mh/_private/plugin.py
```diff
@@ -549,6 +561,38 @@ def _can_run_test(self, item: pytest.Item, data: MultihostItemData | None) -> bo
...
+
+        if preferred_topology != current_topology:
+            self.logger.info(
+                f"{item.location[0]}::{item.location[2]} "
+                f"Preferred topology set to ({preferred_topology}), deselecting test."
+            )
+            return False
+
+        self.logger.info(f"{item.location[0]}::{item.location[2]} ")
+        return True
```
pbrezina, line 594: I would remove the logs completely, it just taints the output.

pbrezina (changes requested):

pytest_mh/_private/plugin.py
danlavu: I didn't think about that, hrmm.... I intended it for debugging, ad hoc runs, I suppose you would disable preferred topologies and just set the topology marker via the cli to get the result I want. Okay, I'll remove it.

docs/articles/tips-and-tricks/preferred-topology.rst
thalman, line 6: > where a single topology is adequate coverage

I would like to understand how this differs from `--mh-topology=<prefered topology>`. Could you explain that also in the documentation, not only in reply to my comment? I think that I get that but I'm not sure so more detailed description of the difference might be useful for others too.

danlavu:
example tests
```
@pytest.mark.topology(KnownTopologyGroup.AnyProvider)
@pytest.mark.preferred_topology(KnownTopology.IPA)
def test_preferred_mark_known_topology_ipa():
    pass


@pytest.mark.topology(KnownTopologyGroup.AnyProvider)
@pytest.mark.preferred_topology("ipa")
def test_preferred_mark_string_ipa():
    pass


@pytest.mark.topology(KnownTopologyGroup.AnyAD)
@pytest.mark.preferred_topology("samba")
def test_preferred_mark_samba_any_ad_topology_group():
    pass


@pytest.mark.topology(KnownTopologyGroup.AnyProvider)
@pytest.mark.preferred_topology("ldap")
def test_preferred_mark_string_ldap():
    pass


@pytest.mark.topology(KnownTopology.Samba)
@pytest.mark.preferred_topology(KnownTopology.LDAP)
def test_preferred_mark_ldap_invalid_topology_mark():
    pass


@pytest.mark.topology(KnownTopologyGroup.AnyProvider)
@pytest.mark.importance("high")
def test_any_provider():
    pass
```

example output
```
collected 19 items / 11 deselected / 8 selected

tests/preferred_mark_test.py::test_any_provider (ad) 
tests/preferred_mark_test.py::test_preferred_mark_known_topology_ipa (ipa) 
tests/preferred_mark_test.py::test_preferred_mark_string_ipa (ipa) 
tests/preferred_mark_test.py::test_any_provider (ipa) 
tests/preferred_mark_test.py::test_preferred_mark_string_ldap (ldap) 
tests/preferred_mark_test.py::test_any_provider (ldap) 
tests/preferred_mark_test.py::test_preferred_mark_samba_any_ad_topology_group (samba) 
tests/preferred_mark_test.py::test_any_provider (samba) 
```

thalman (approved):
LGTM, thanks for the work

jakub-vavra-cz (approved):
Looks great. Thank You.

pytest_mh/_private/plugin.py
```diff
@@ -521,6 +523,27 @@ def _create_logger(self, verbose) -> logging.Logger:
...
+        return item.get_closest_marker(name="preferred_topology") is not None
+
+    def _run_preferred_topology(self, mark: pytest.Mark, current_topology: str) -> bool:
+        if not mark.args:
+            raise ValueError
+
+        arg = mark.args[0]
+
+        if isinstance(arg, (TopologyMark, KnownTopologyBase)):
+            arg = arg.name.lower()
```
pbrezina, line 536: This is wrong and works only by chance. `KnownTopologyBase` is `enum` and not `TopologyMark`. `enum` has `name` and `value` attribute, but the `name` can be different from the topology name...

```
(Pdb) l
531  	            raise ValueError
532  	
533  	        arg = mark.args[0]
534  	
535  	        if isinstance(arg, (TopologyMark, KnownTopologyBase)):
536  ->	            breakpoint()
537  	            arg = arg.name.lower()
538  	        elif isinstance(arg, str):
539  	            arg = arg.lower()
540  	        else:
541  	            raise TypeError("Preferred typology is the wrong type!")
(Pdb) p type(arg)
<enum 'KnownTopology'>
(Pdb) arg
<KnownTopology.LDAP: <sssd_test_framework.config.SSSDTopologyMark object at 0x7f2e5221ce10>>
(Pdb) arg.value
<sssd_test_framework.config.SSSDTopologyMark object at 0x7f2e5221ce10>
(Pdb) arg.name
'LDAP'
(Pdb) arg.value.name
'ldap'
(Pdb) type(arg.value)
<class 'sssd_test_framework.config.SSSDTopologyMark'>

```

The code should be

```python
    def _can_run_preferred_topology(self, mark: pytest.Mark, current_topology: str) -> bool:
        if not mark.args or len(mark.args) > 1:
            raise ValueError("Unexpected number of arguments to pytest.mark.preferred_topology")

        arg = mark.args[0]

        if isinstance(arg, KnownTopologyBase):
            name = arg.value.name
        if isinstance(arg, TopologyMark):
            name = arg.name
        elif isinstance(arg, str):
            name = arg
        else:
            raise TypeError("Preferred typology is the wrong type!")

        if name == current_topology:
            return True

        return False
```

pbrezina (changes requested):

pytest_mh/_private/plugin.py
```diff
@@ -547,6 +570,16 @@ def _can_run_test(self, item: pytest.Item, data: MultihostItemData | None) -> bo
...
 
+        if (
+            self._has_preferred_topology(item)
+            and self.mh_ignore_preferred_topology is False
+            and data.topology_mark is not None
+        ):
+            current_topology = data.topology_mark.name
+            preferred_topology = item.get_closest_marker(name="preferred_topology")
+            if preferred_topology:
+                return self._run_preferred_topology(preferred_topology, current_topology)
```
pbrezina, line 581: This ignores the case where --mh-topology is set and `self._has_preferred_topology(item)` is actually checked twice (the second call is on line 580). You can rewrite it to:

```python
        # Run only for preferred topology unless specific topology is requested or the marker is ignored
        if not self.mh_topology and not self.mh_ignore_preferred_topology:
            preferred_topology = item.get_closest_marker(name="preferred_topology")
            if preferred_topology is not None and data.topology_mark is not None:
                return self._run_preferred_topology(preferred_topology, data.topology_mark.name)
```

pbrezina (changes requested):

pytest_mh/_private/plugin.py
danlavu: @pbrezina I'm testing the code, and it's failing to hit the condition. When I check the type being passed in, it's ```<enum 'KnownTopology'>```. I can change it to  ``` instance(arg, enum) ``` and it'll work. Where is KnownTopology coming from? Looking at the class, I don't see a KnownTopology member like KnownTopologyBase.

pbrezina: KnownTopology is defined in the sssd-test-framework, it inherits from KnownTopologyBase. `isinstance(arg, KnownTopologyBase)` works:

```
@pytest.mark.preferred_topology(KnownTopology.LDAP)
@pytest.mark.topology(KnownTopologyGroup.AnyProvider)
def test_preferred(client: Client):
    assert True

> /var/home/pbrezina/workspace/pytest-mh/pytest_mh/_private/plugin.py(533)_can_run_preferred_topology()
-> breakpoint()
(Pdb) l
528  	            raise ValueError("Unexpected number of arguments to pytest.mark.preferred_topology")
529  	
530  	        arg = mark.args[0]
531  	
532  	        if isinstance(arg, KnownTopologyBase):
533  ->	            breakpoint()
534  	            name = arg.value.name
535  	        if isinstance(arg, TopologyMark):
536  	            name = arg.name
537  	        elif isinstance(arg, str):
538  	            name = arg
```

I assume you think it did not hit the condition because you see "Preferred typology is the wrong type!" exception... you need to replace second if with elif, of course.

danlavu: *facepalm*

pbrezina (approved):

PR #100, opened 2025-02-27
Workaround timeout issue with SSH_AUTH_AGAIN

justin-stephenson:
When the timeout is set to 1 in LibsshSession(timeout=1), pylibssh no longer returns an exception for a wrong password SSH connect()

    with pytest.raises(SSHAuthenticationError):
        client.ssh("user1", "Wrong").connect()

Increase the minimum timeout a couple seconds here.

Regarding 'LibsshSession(timeout=1)', a very low value must be set here, to allow the timeout set in

        client.host.conn.run(..., timeout=X)

to work as expected. This is due to code becoming blocked in libssh which is C not Python, but the Python signal is delayed until we get back to the Python code

### DIFF
--- a/pytest_mh/conn/ssh.py
+++ b/pytest_mh/conn/ssh.py
@@ -454,7 +454,7 @@ class SSHClient(Connection[SSHProcess, SSHProcessResult]):
         # since Python will not deliver signal if the code is blocked in C
         # library. The signal is deliver only after we get back to the Python
         # code.
-        self.__conn: LibsshSession = LibsshSession(timeout=1)
+        self.__conn: LibsshSession = LibsshSession()
 
     def _read_private_key(
         self,
@@ -518,6 +518,7 @@ class SSHClient(Connection[SSHProcess, SSHProcessResult]):
                 port=self.port,
                 host_key_checking=False,
             )
+            self.__conn.set_ssh_options("timeout", 1)
         except LibsshSessionException as e:
             raise SSHAuthenticationError(self.host, self.port, self.user, e.message)
 


pytest_mh/conn/ssh.py
```diff
@@ -432,7 +432,7 @@ def __init__(
         :param logger: Multihost logger.
         :type logger: MultihostLogger
         :param timeout: Timeout in seconds (defaults to 300), value
-            ``0`` means that timeout is disabled.
+            ``0`` means that timeout is disabled. 3 second minimum value.
```
ikerexxe, line 435: If 3 is the minimum value, then add some code to check this argument and increment it (or error) in case someone sets a lower value.

pytest_mh/conn/ssh.py
```diff
@@ -454,7 +454,10 @@ def __init__(
         # since Python will not deliver signal if the code is blocked in C
         # library. The signal is deliver only after we get back to the Python
         # code.
-        self.__conn: LibsshSession = LibsshSession(timeout=1)
+        # This must be set to a low value to allow timeout X value as
+        # shown below to work properly
+        # client.host.conn.run(..., timeout=X)
```
ikerexxe, line 459: I do not understand this comment followed by the previous one. Can you check it?

ikerexxe (commented):
In addition to the two inline comments. Can you add the following to the commit message?
```
Fixes: 9d8c93dc10fa ("conn: add support for timeout")

**Comment by pbrezina** [comment]
3 seconds did not work for me, we can not use any arbitrary value. But we can workaround it:

```
diff --git a/pytest_mh/conn/ssh.py b/pytest_mh/conn/ssh.py
index 9a8a36a..777a2eb 100644
--- a/pytest_mh/conn/ssh.py
+++ b/pytest_mh/conn/ssh.py
@@ -454,7 +454,7 @@ class SSHClient(Connection[SSHProcess, SSHProcessResult]):
         # since Python will not deliver signal if the code is blocked in C
         # library. The signal is deliver only after we get back to the Python
         # code.
-        self.__conn: LibsshSession = LibsshSession(timeout=1)
+        self.__conn: LibsshSession = LibsshSession()
 
     def _read_private_key(
         self,
@@ -518,6 +518,7 @@ class SSHClient(Connection[SSHProcess, SSHProcessResult]):
                 port=self.port,
                 host_key_checking=False,
             )
+            self.__conn.set_ssh_options("timeout", 1)
         except LibsshSessionException as e:
             raise SSHAuthenticationError(self.host, self.port, self.user, e.message)
```

**Review by pbrezina — CHANGES_REQUESTED** [review]
See my comment above.

It would be also nice if you could send PR to pylibssh to handle this case correctly. It will take some time to get a release, so we still need a workaround.

* https://api.libssh.org/master/group__libssh__auth.html#ga50c0c150f8c4703e7ee49b3e3e3ca215 SSH_AUTH_AGAIN: In nonblocking mode, you've got to call this again later.
* https://github.com/ansible/pylibssh/blob/devel/src/pylibsshext/session.pyx#L446

**Comment by pbrezina** [comment]
> > See my comment above.
> > It would be also nice if you could send PR to pylibssh to handle this case correctly. It will take some time to get a release, so we still need a workaround.
> 
> Do you mean with a fix in pylibssh pytest-mh timeouts should work properly with `self.__conn: LibsshSession = LibsshSession()` or `self.__conn: LibsshSession = LibsshSession(timeout=1)` ???

With `self.__conn: LibsshSession = LibsshSession(timeout=1)`.

> Also, in my testing I tried with `self.__conn: LibsshSession = LibsshSession(timeout=10)` and the below test times out in 10 seconds as expected. Is there something missing in this approach of setting the same value provided from the test (or 300 default if not provided)?
> 
> ```
> @pytest.mark.topology(KnownTopology.Client)
> def test_timeout(client: Client):
>     result = client.host.conn.run(
>         """
>         echo 'stdout before';
>         >&2 echo 'stderr before';
>         sleep 30;
>         echo 'stdout after';
>         >&2 echo 'stderr after'
>         """,
>         timeout=10,
>     )
> ```

Timeout in `self.__conn: LibsshSession = LibsshSession(timeout=10)` is the minimal functional timeout. So `client.host.conn.run("sleep 20", timeout=15)` will timeout in 15 seconds, `client.host.conn.run("sleep 20", timeout=10)`  will timeout in 10 seconds but `client.host.conn.run("sleep 20", timeout=5)` will still timeout in 10 seconds which is not expected.

**Comment by justin-stephenson** [comment]
> > > See my comment above.
> > > It would be also nice if you could send PR to pylibssh to handle this case correctly. It will take some time to get a release, so we still need a workaround.
> > 
> > 
> > Do you mean with a fix in pylibssh pytest-mh timeouts should work properly with `self.__conn: LibsshSession = LibsshSession()` or `self.__conn: LibsshSession = LibsshSession(timeout=1)` ???
> 
> With `self.__conn: LibsshSession = LibsshSession(timeout=1)`.
> 

Is this what you had in mind? https://github.com/justin-stephenson/pylibssh/commit/a51580976fbe78a3134d5b1ea38b90560fca6273

It works in my initial testing.

**Comment by pbrezina** [comment]
No, not really.

FIrst things first, lets finish this PR so we can unblock PR CI. Because who knows how long we will be wait for next pylibssh release... that is pretty much just this diff: https://github.com/next-actions/pytest-mh/pull/100#issuecomment-2694071931

For pylibssh fix, it should be something like this :

```
do:
    rc = libssh.ssh_userauth_password(self._libssh_session, NULL, password.encode())
while rc == libssh.SSH_AUTH_AGAIN
```

justin-stephenson:
> No, not really.
> 
> FIrst things first, lets finish this PR so we can unblock PR CI. Because who knows how long we will be wait for next pylibssh release... that is pretty much just this diff: [#100 (comment)](https://github.com/next-actions/pytest-mh/pull/100#issuecomment-2694071931)

Updated, I re-worded the commit message please check it.
> 
> For pylibssh fix, it should be something like this :
> 
> ```
> do:
>     rc = libssh.ssh_userauth_password(self._libssh_session, NULL, password.encode())
> while rc == libssh.SSH_AUTH_AGAIN
> ```

Okay thanks I'll try this though it looks like there is no handling of `SSH_AUTH_AGAIN` at all in pylibssh

justin-stephenson:
Updated PR title.

ikerexxe:
Let me see if I understand the problem. We can't create an object of type `LibsshSession` with a timeout of 1 because pylibssh returns a value of SSH_AUTH_AGAIN, but this can't be managed by `LibsshSession`. But we can set this value later on through `set_ssh_options()`. Am I right?

ikerexxe (approved):
> > Let me see if I understand the problem. We can't create an object of type `LibsshSession` with a timeout of 1 because pylibssh returns a value of SSH_AUTH_AGAIN, but this can't be managed by `LibsshSession`. But we can set this value later on through `set_ssh_options()`. Am I right?
> 
> With the `LibsshSession(timeout=1)` we see libssh returns SSH_AUTH_AGAIN but pylibssh `authenticate_password` returns success rc == 0 in this case. Moving setting the timeout after the connect() call works around this.

Good. Thank you for taking the time to explain. LGTM!

pbrezina (approved):

PR #101, opened 2025-04-24
utils: SELinux file context

ikerexxe:
Implement SELinux file context class and LinuxFileSystem method.


pytest_mh/utils/fs.py
```diff
@@ -230,6 +231,23 @@ def exists(self, path: str) -> bool:
 
         return False
 
+    def selinux_context(self, path: str) -> SELinuxContext | None:
```
jakub-vavra-cz, line 234: If this can be None, we will always need to run two asserts
assert x in not None, "Could not detect selinux context"
assert x.type == "foo_t", "Context does not match"
 
I would rather see the object of the type SELinuxContext that has "bogus values" that You can just directly compare instead.

pytest_mh/utils/selinux.py
```diff
@@ -0,0 +1,72 @@
...
+        return (
+            self.user == other.user
+            and self.role == other.role
+            and self.type == other.type
+            and self.sensitivity == other.sensitivity
+            and self.categories == other.categories
+        )
+
+    @classmethod
+    def FromOutput(cls, stdout: str) -> SELinuxContext | None:
```
jakub-vavra-cz, line 61:
```suggestion
    def FromOutput(cls, stdout: str) -> SELinuxContext:
```

pytest_mh/utils/fs.py
```diff
@@ -230,6 +231,23 @@ def exists(self, path: str) -> bool:
...
+        :type path: str
+        :return: SELinux file context
+        :rtype: SELinuxContext or None
+        """
+        self.logger.info(f'Checking SELinux context for "{path}"')
+        result = self.host.conn.exec(["ls", "-Z", path], log_level=ProcessLogLevel.Error, raise_on_error=False)
+
+        if result.rc != 0:
+            return None
+
```
jakub-vavra-cz, line 248:
```suggestion
```

jakub-vavra-cz (commented):
See inline.

pytest_mh/utils/fs.py
ikerexxe: Returning both SELinuxContext or None is generally preferred in Python for its clarity and flexibility in handling the potential absence of a value. It could happen that the command didn't run successfully and we don't get any SELinux context, thus checking for it is the only way of getting the proper verbosity output in the logs. I could also return an exception, but that wouldn't be as clear as None.

pytest_mh/utils/selinux.py
ikerexxe: Fixed

pytest_mh/utils/selinux.py
```diff
@@ -0,0 +1,73 @@
...
+    def FromOutput(cls, stdout: str) -> SELinuxContext:
+        match = re.match(r"([^ ]+)\s+", stdout)
+
+        if not match:
+            raise ValueError("Unexpected value: expecting space separated string")
+
+        selinux_context = match.group(1)
+        labels = selinux_context.split(":")
+
+        if len(labels) != 5:
```
alexey-tikhonov, line 70:
```
$ ls -Z sssd_ssh 
unconfined_u:object_r:user_home_t:s0 sssd_ssh
```

Are `categories` really always present?

alexey-tikhonov:
How is this going to be used?

ikerexxe:
This will be used by shadow project to check the correct labeling of SELinux files. More specifically the issue that triggered this development is https://github.com/shadow-maint/shadow/issues/940.

I thought this might be useful for other projects that use pytest-mh and its file processing utilities, and that's why I decided to implement it in pytest-mh instead of in the shadow test framework.

pytest_mh/utils/fs.py
```diff
@@ -230,6 +231,23 @@ def exists(self, path: str) -> bool:
 
         return False
 
+    def selinux_context(self, path: str) -> SELinuxContext | None:
+        """
+        Checks SELinux file context.
```
alexey-tikhonov, line 236: Maybe 'Reads' or 'Gets'? 'Checks' is misleading here, imo.

pytest_mh/utils/fs.py
```diff
@@ -230,6 +231,23 @@ def exists(self, path: str) -> bool:
...
+
+        :param path: File path.
+        :type path: str
+        :return: SELinux file context
+        :rtype: SELinuxContext or None
+        """
+        self.logger.info(f'Checking SELinux context for "{path}"')
```
alexey-tikhonov, line 243: The same - 'Reading'.

pytest_mh/utils/selinux.py
alexey-tikhonov: What will `ls -Z` return if selinux is completely disabled?

ikerexxe: `SELinuxContext` will be `None`. The test case should check for the SELinux feature to be enabled to run this test (either completely or partially).

pytest_mh/utils/fs.py
ikerexxe: Changed to 'Gets'

pytest_mh/utils/fs.py
ikerexxe: I prefer 'Gets' and 'Getting', so I updated it accordingly

pytest_mh/utils/selinux.py
alexey-tikhonov: Do you have `ls -Z` output example in this case?

alexey-tikhonov:
```
       -Z, --context
              print any security context of each file
```
What about other types of "security context"?
I don't know if stuff like 'AppArmor' will show up here is well?
Or is this specifically about "selinux"?

alexey-tikhonov: > The test case should check for the SELinux feature to be enabled

Or maybe `selinux_context()` could check this and return 'None' if disabled?

ikerexxe: > Or is this specifically about "selinux"?

This is just about SELinux context. I don't know how AppArmor works and from what I heard Suse is moving away from it to SELinux.

> Or maybe selinux_context() could check this and return 'None' if disabled?

There, there is no way of knowing whether SELinux is enabled or SELinux labeling failed. We need a way to distinguish those cases during test execution. In this particular case the difference between these two cases is important because the shadow CI runs on Fedora, Debian, Alpine and OpenSuse. Some of these distributions do not provide SELinux.

ikerexxe: I don't know if I inadvertently resolved this conversation, so I'm reopening it. Feel free to resolve it if I clarified all your doubts.

alexey-tikhonov: I don't see a "resolve" button, likely lack of access rights.

pbrezina (commented):
How about creating standalone SELinux utility and implementing the method there instead of mixing this in fs.py?

ikerexxe:
I'm fine with that, but apart from checking SELinux file context labels I wouldn't like to implement anything else for now. Would you be okay with such a simple class?

pytest_mh/utils/selinux.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def __str__(self) -> str:
+        return f"({self.user}:{self.role}:{self.type}:{self.sensitivity}:{self.categories})"
+
+    def __repr__(self) -> str:
+        return str(self)
+
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, SELinuxContext):
+            raise TypeError(f"Unexpected type: {type(other)}, expecting SELinuxContext")
```
jakub-vavra-cz, line 55: Comparing with other type means not equal.
If I want to compare given context for example to None this way it will work.
```suggestion
        if not isinstance(other, SELinuxContext):
            return False
```

jakub-vavra-cz (changes requested):
See inline.
I do not want exception when comparing the SELinuxContext to None.

pytest_mh/utils/selinux.py
ikerexxe: Makes sense. Updated.

jakub-vavra-cz (approved):

PR #104, opened 2025-07-02
.github: remove workaround to install containernetworking-plugins

ikerexxe:


ikerexxe:
@pbrezina same as https://github.com/SSSD/sssd-ci-containers/pull/137 but for pytest-mh

pbrezina (approved):

PR #106, opened 2025-08-20
Adding utility for hostname management

shridhargadekar:
Hostnamectl utility manages the hostname
configurations of host.

### DIFF
--- a/pytest_mh/utils/hostname.py
+++ b/pytest_mh/utils/hostname.py
@@ -1,0 +1,101 @@
+"""Manage system hostname using the 'hostname' command."""
+
+from __future__ import annotations
+
+from pytest_mh import MultihostHost, MultihostUtility
+from pytest_mh.conn import ProcessResult
+
+__all__ = ["HostnameUtils"]
+
+
+class HostnameUtils(MultihostUtility[MultihostHost]):
+    """
+    Provides a Python wrapper for the basic 'hostname' command-line utility.
+    """
+
+    def __init__(self, host: MultihostHost) -> None:
+        """
+        Initialize the HostnameUtils.
+
+        :param host: The multihost host instance.
+        :type host: MultihostHost
+        """
+        super().__init__(host)
+
+    @property
+    def name(self) -> str:
+        """
+        Gets the current system hostname by running `hostname`.
+
+        :return: The current hostname as a string.
+        :rtype: str
+        """
+        result = self.host.conn.exec(["hostname"])
+        return result.stdout.strip()
+
+    @property
+    def short_name(self) -> str:
+        """
+        Gets the short host name (without the domain) by running `hostname -s`.
+
+        :return: The short hostname as a string.
+        :rtype: str
+        """
+        result = self.host.conn.exec(["hostname", "-s"])
+        return result.stdout.strip()
+
+    @property
+    def fqdn(self) -> str:
+        """
+        Gets the Fully Qualified Domain Name (FQDN) by running `hostname -f`.
+
+        :return: The FQDN as a string.
+        :rtype: str
+        """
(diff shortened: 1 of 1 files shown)


shridhargadekar:
Containerized environment does not allow the modification of a hostname

shridhargadekar:
The use case: 
`adcli`, `realmd` should be able to join the clients with modified hostname. Hostname with different character-classes. There are multiple scenarios involving hostname.
`ad-gpo` also needs the hostname modification for some of tests.

danlavu:
ad dynamic dns needs this. The IP stuff, please look at https://github.com/SSSD/sssd-test-framework/pull/197 

There is an overlap in getting network information.

pytest_mh/utils/hostname.py
```diff
@@ -0,0 +1,101 @@
...
+        Gets the primary network address for the hostname by running `hostname -i`.
+
+        :return: The primary IP address as a string.
+        :rtype: str
+        """
+        result = self.host.conn.exec(["hostname", "-i"])
+        return result.stdout.strip()
+
+    @property
+    def all_ip_addresses(self) -> list[str]:
```
danlavu, line 81: We can easily add get all IPs to the IPUtils but I don't think it will be used that much or at all. What scenario will use this?

pytest_mh/utils/hostname.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        Initialize the HostnameUtils.
+
+        :param host: The multihost host instance.
+        :type host: MultihostHost
+        """
+        super().__init__(host)
+
+    @property
+    def name(self) -> str:
```
danlavu, line 26: I think this should be part of the client role. So it's easier to access when writing tests. 

```client.hostname```

File: pytest_mh/utils/hostname.py [hunk 3]
```diff
@@ -0,0 +1,101 @@
...
+        Gets the system's domain name by running `hostname -d`.
+
+        :return: The domain name as a string.
+        :rtype: str
+        """
+        result = self.host.conn.exec(["hostname", "-d"])
+        return result.stdout.strip()
+
+    @property
+    def ip_address(self) -> str:
```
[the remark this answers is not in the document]
shridhargadekar: This PR provides utilities provided by `hostname`.  
The location to keep this utility has been already discussed. As this utility is expected to be used by multiple projects, we agreed to keep it here. 
The overlapping of utilities are expected.

pytest_mh/utils/hostname.py
```diff
@@ -0,0 +1,101 @@
...
+        Gets the short host name (without the domain) by running `hostname -s`.
+
+        :return: The short hostname as a string.
+        :rtype: str
+        """
+        result = self.host.conn.exec(["hostname", "-s"])
+        return result.stdout.strip()
+
+    @property
+    def fqdn(self) -> str:
```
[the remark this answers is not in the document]
shridhargadekar: Already answered.

pytest_mh/utils/hostname.py
shridhargadekar: Check previous reply.

pytest_mh/utils/hostname.py
shridhargadekar: Check previous reply

danlavu:
It is there, just as one method.

pbrezina (commented):
Since you allow to set the name, a proper setup and teardown methods must be implemented.

Otherwise, looks good to me. I don't have an strong opinion on whether to do it as properties or as Dan suggests, you should talk it out. But since the purpose of this utility is to simplify something, then the properties looks easier to use.

PR #111, opened 2025-11-19
Add file check to auditd for log files

spoore1:
In some cases, log files are cleaned up before the auditd teardown can
perform it's cleanup.   Adding a check to see if the files exist before
running restore and cleanup.  If no files found, exit and skip.

### DIFF
--- a/pytest_mh/utils/auditd.py
+++ b/pytest_mh/utils/auditd.py
@@ -89,10 +89,13 @@ class Auditd(MultihostUtility[MultihostHost]):
                 f"""
                 set -e
 
-                for f in "{self._backup}"/audit/audit.log*; do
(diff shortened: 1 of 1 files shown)


pbrezina:
AFAIK -f "file*" literally checks for "file*" presence (it does not evaluate the wildcard), so I mean something like this

```
path="{self._backup}"/audit
if [ -d "$path" ] && [ -n "$(ls -A "$path)" ]; then
```

ie folder exist and is not empty

pbrezina (changes requested):
@spoore1 Thank you. Can you change the commit subject to `utils: ...`?

Also, can you open sssd pull request and pull this pytest-mh branch so we can see that it solves the issue?

PR #112, opened 2026-03-31
ssh: set ansible-pylibssh logging to INFO

spoore1:
ansible-pylibssh version 1.4.0 defaults to debug logging to the main pytest logging.  This is causing excessive log messages in the pytest run log.   This is to limit the logging.


justin-stephenson (approved):
LGTM, thanks.

danlavu (approved):

PR #113, opened 2026-04-22
ssh: disable ansible-pylibssh logging

spoore1:
the pylibssh logging appears to still be too verbose in some
environments.   Disabling it entirely to avoid this excessive logging
that is making troubleshooting normal test failures more difficult.


spoore1:
I think this is now working.   Tests run with this version didn't flood the logs with pylibssh messages.

We'll need to make sure we have the tag in place upstream when it's merged so the CI jobs pick up the correct version.

PR #114, opened 2026-05-20
tests: cli builder supports list in VALUE type argument

pbrezina:


pytest_mh/cli.py
```diff
@@ -142,6 +151,11 @@ def __match_shell(self, shell: Type[Shell]):
 
     def __build(self, command: str | None, args: CLIBuilderArgs, quote_value: bool) -> list[str]:
         def _get_option(name: str) -> str:
+            # Return option name as is, if it already starts with dash
+            # This adds support for "-arg" type tools and short args "-a"
+            if (name.startswith("-")):
```
sumit-bose: Hi,

it looks like `black` would prefer
```
    if name.startswith("-"):
```

It looks like there is an CI issue running `black`, I'm not sure if this is currently expected here or if this needs fixing?

bye,
Sumit

pbrezina: Haha, fixed. C mindset kicked in, we don't use parentheses in Python of course in this context :-)

sumit-bose: Thanks.

sumit-bose (approved):
Hi,

patch works for me, ACK.

bye,
Sumit

pytest_mh/cli.py
pbrezina: Fixed.

PR #115, opened 2026-06-18
adding optional parameter to host.conn.ssh to remove carriage returns in powershell

danlavu:


danlavu:
I don't know if this should go here. I think it's cleaner than adding it to the framework as a parser function. All of the powershell output needs to be cleaned before they can be used, like in the example below. 

```
        records = self.host.conn.run(f"dig +short +norecurse {name} '@{self.server}'", ps=True).stdout_lines
        zones = self.list_zones()

        if records is not None:
            for zone in zones:
                if zone in name:
```

spoore1 (changes requested):
Can the lint failures from tox checks be resolved?   I'm guessing you'll need to rerun black locally as described?

danlavu:
@spoore1, the tests host has a new interpreter, and 3.11 was listed. I am updating the tox configuration to include 3.14, 3.15, and skipping missing interpreters.

spoore1 (approved):
LGTM

PR #116, opened 2026-07-13
utils: add comprehensive stat method using jc parser

ikerexxe:


ikerexxe:
The changes implemented in this PR are used in the tests developed at https://github.com/shadow-maint/shadow/pull/1673

pytest_mh/utils/fs.py
```diff
@@ -757,3 +870,17 @@ def sed(self, command: str, path: str, args: list[str] | None = None) -> Process
...
+        """
+        Get file status information.
+
+        :param path: File path.
+        :type path: str
+        :return: File status information.
+        :rtype: StatEntry
+        """
+        self.logger.info(f'Getting file status for "{path}"')
+        result = self.host.conn.exec(["stat", "-c", "%a:%h:%U:%G:%s:%Y", path], log_level=ProcessLogLevel.Error)
```
pbrezina: How about using jc? https://kellyjonbrazil.github.io/jc/docs/parsers/stat We can add it as a dependency.

It would allow to parse full stat information, not just this limited set.

ikerexxe: Thanks for the hint. I've updated the PR with this

pbrezina (approved):
Thanks. I did not test it, but the patch is simple. If it works in your tests, we can merge it.

ikerexxe:
Yes, it does work so let's merge it

PR #119, opened 2026-08-04
fix: Use base64 for transport in conn to avoid quote escape hell.

jakub-vavra-cz:
Quotes are being lost between python/cygwin/powershell resulting in failures like this one:

INTERNALERROR>     # Backup computers
INTERNALERROR>     $backup_computers = Join-Path $tmpdir computers.txt
INTERNALERROR>     $result = Get-ADObject -SearchBase "cn=computers,$basedn" -LDAPFilter "(objectClass=computer)"
INTERNALERROR>     foreach ($r in $result) { $r.DistinguishedName | Add-Content -Path $backup_computers }
INTERNALERROR> 
INTERNALERROR>     Write-Output $tmpdir.FullName
INTERNALERROR>   CWD:
INTERNALERROR>   Env:
INTERNALERROR>   Output:
INTERNALERROR>   Error output:
INTERNALERROR>     At line:1 char:19
INTERNALERROR>     + $basedn = DC=child,DC=ad-ed1r,DC=test
INTERNALERROR>     +                   ~
INTERNALERROR>     Missing argument in parameter list.
INTERNALERROR>     At line:17 char:33
INTERNALERROR>     + | Where-Object { $_.ZoneType -eq \"Primary\" -and $_.IsAutoCreated -e ...
INTERNALERROR>     +                                 ~
INTERNALERROR>     You must provide a value expression following the '-eq' operator.
INTERNALERROR>     At line:17 char:34
INTERNALERROR>     + | Where-Object { $_.ZoneType -eq \"Primary\" -and $_.IsAutoCreated -e ...
INTERNALERROR>     +                                  ~~~~~~~~~~~
INTERNALERROR>     Unexpected token '\"Primary\"' in expression or statement.
INTERNALERROR>         + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
INTERNALERROR>         + FullyQualifiedErrorId : MissingArgument

Use base64 encoded commands instead.

### DIFF
--- a/pytest_mh/conn/__init__.py
+++ b/pytest_mh/conn/__init__.py
@@ -1,10 +1,13 @@
 from __future__ import annotations
 
+import base64
 import itertools
 import os
+import re
 import shlex
 import signal
 import textwrap
+import xml.etree.ElementTree as ET
 from abc import ABC, abstractmethod
 from enum import Enum, auto
 from typing import TYPE_CHECKING, Any, Generator, Generic, NoReturn, Self, TypeVar
@@ -1149,6 +1152,10 @@ class Shell(ABC):
         self.shell_command: str = shell_command
         """Shell command that will execute user scripts."""
 
+    def sanitize_stderr(self, stderr: list[str]) -> list[str]:
+        """Return stderr lines; subclasses may strip shell-specific noise."""
+        return stderr
+
     @abstractmethod
     def build_command_line(self, script: str, *, cwd: str | None, env: dict[str, Any]) -> str:
         """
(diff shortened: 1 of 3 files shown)


pbrezina:
Can you be more explicit what is the problem? And why XML is involved?

jakub-vavra-cz:
> Can you be more explicit what is the problem? And why XML is involved?
Quotes are lost on the way in some cases between python/bash/cygwin/powershell breaking the commands like above.
The fix is to use "-EncodedCommand" with all of the parts encoded as base64 instead of trying to escape/quote.
Unfortunately the the output is returned wrapped in xml in this case so need to strip it to the the usual plain text outputs.

pbrezina:
But why? Why now? Why is it malformed? What cases?

You mention bash, but changed only powershell - so does it happen only if you run bash inside powershell?

The error snippet from PR description executes code that we rely on for two years now. So what changed that it fails suddenly? Perhaps the failure is caused by https://github.com/next-actions/pytest-mh/commit/84c080ee0562aad71d067287d584c78b4f1c7549?

jakub-vavra-cz:
I hit the issue above when working on ADForest topology:
https://github.com/SSSD/sssd-test-framework/pull/263/changes
The issue seems unrelated to 84c080e.

spoore1 (commented):
So far in testing and from a quick review of the code, it looks good.   I ran sssd system ad tests with this.   While I did see 1 unexpected failure, that failure appears to have been a flaky test.   The rest all passed.

I was able to see the test updates run and pass also in the tox pipeline job.

spoore1 (approved):
While I can't reproduce the issue, I have run this with existing AD tests and they're passing:

```
198 passed, 1 skipped, 786 deselected in 3548.58s (0:59:08)
```

pbrezina:
I still would like to see a reproducer and answers to the questions "But why? Why now? Why is it malformed? What cases?".

The error snippet from description is from backup/restore code that we use for a long time. This is not really trivial fix, it changes the way data are transferred. Perhaps correct and better, but I want to understand the problem before merging it.

spoore1:
Requesting review again to wait for a way to reproduce.

alexey-tikhonov:
Issue was solved in a different way meanwhile, closing this PR.

